PR #120, opened 2024-09-26
improve how we handle rpc requests for user tokens. Log out informati…

moose-code:
…on when cache isn't hit


src/Erc20.ts
```diff
@@ -12,49 +12,106 @@ const contractABI = require("../abis/ERC20.json");
...
-      erc20token.methods.decimals().call(),
-      erc20token.methods.symbol().call(),
+      erc20token.methods
+        .name()
+        .call()
+        .catch((e) => {
+          console.error(
+            `[getErc20TokenDetails] Error fetching name: ${e.message}`
+          );
+          return "";
```
moose-code, line 61: I believe some meme coin decided to break the ERC20 status quo and do something custom and dogey with on e of these methods and that probably broke things

PR #125, opened 2024-09-27
Refactor Event Handlers

jfarid27:
# Summary

- Refactors Event Handlers into separate files for easier readability and maintenance.
- Moves all dates into datetime objects.

# Review

- [ ] Check if the datetime conversion is correct in Interval Snapshots and the ID correctly normalizes.


src/IntervalSnapshots.ts
```diff
@@ -112,14 +111,19 @@ function getNumberOfSecondsInInterval(interval: SnapshotInterval): bigint {
   }
 }
 
-// Function to return the interval ID for a given snapshot entity
 function getIdForEntityByInterval(
```
jfarid27, line 114: @moose-code My only question is if this update works. Previously with the `lastUpdatedTimestamp` being a bigint, there was this computation being done on intervals. I'm wondering if updating this code to convert the date into a unix time and continuing the same computation will work.

src/IntervalSnapshots.ts
```diff
@@ -112,14 +111,19 @@ function getNumberOfSecondsInInterval(interval: SnapshotInterval): bigint {
...
-  lastUpdatedTimestamp: bigint,
+  lastUpdatedTimestamp: Date,
   numberOfSeconds: bigint
 ): string {
-  let interval = lastUpdatedTimestamp / numberOfSeconds;
+  // Convert Date to Unix timestamp in seconds
+  let lastUpdatedTimestampInSeconds = Math.floor(lastUpdatedTimestamp.getTime() / 1000);
+
+  // Perform interval calculation
+  let interval = BigInt(lastUpdatedTimestampInSeconds) / numberOfSeconds;
```
moose-code, line 123: Looks good! Also running fine locally

moose-code (approved):
Nice, LGTM 💪

PR #145, opened 2024-10-15
Adding loaders to improve historic sync speed

moose-code:
This code now makes use of loaders: https://docs.envio.dev/docs/HyperIndex/loaders

to improve performance


jfarid27 (commented):
LGTM

PR #172, opened 2024-11-05
Bugfix: Persist Total Liquidity

jfarid27:
# Description

Persist Total Liquidity by keeping value in the pool snapshot.


moose-code:
Nice!!!

PR #175, opened 2024-11-06
Bugfix: Force checksum addresses on gauge cache.

jfarid27:
# Description

Remove lowercasing of addresses and force checksum on gauge addresses.

### DIFF
--- a/src/EventHandlers/Voter.ts
+++ b/src/EventHandlers/Voter.ts
@@ -5,9 +5,9 @@ import {
   Voter_WhitelistToken,
 } from "generated";
 
-import { LiquidityPoolAggregator, Token } from "./../src/Types.gen";
+import { Token } from "./../src/Types.gen";
 import { normalizeTokenAmountTo1e18 } from "./../Helpers";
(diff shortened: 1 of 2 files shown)


src/EventHandlers/Voter.ts
```diff
@@ -59,9 +59,9 @@ Voter.GaugeCreated.handler(async ({ event, context }) => {
   // The pool entity should be created via PoolCreated event from the PoolFactory contract
   // Store pool details in poolRewardAddressStore
   let currentPoolRewardAddressMapping = {
-    poolAddress: event.params.pool,
-    gaugeAddress: event.params.gauge,
-    bribeVotingRewardAddress: event.params.bribeVotingReward,
+    poolAddress: toChecksumAddress(event.params.pool),
+    gaugeAddress: toChecksumAddress(event.params.gauge),
+    bribeVotingRewardAddress: toChecksumAddress(event.params.bribeVotingReward),
```
moose-code, line 64: Wondering the computational cost of all this checksumming, if we've removed the toLowerCase we shouldn't need this now right?

PR #177, opened 2024-11-07
Mode network addition.

jfarid27:
# Description

Adding Mode network.


config.yaml
```diff
@@ -59,6 +59,27 @@ contracts:
     events:
       - event: NotifyReward(address indexed from, uint256 amount)
 networks:
+  - id: 34443 # Mode
+    start_block: 0
```
JasoonS, line 63: I personally prefer including the start block -  But it does probably only save max 3 seconds or so of indexing time.

moose-code (approved):
LGTM 💪

moose-code:
Ran locally and its looking good, just going to also bump version

PR #323, opened 2025-04-17
Some cleanups.

stas:
Mainly removal of hydra and stale/cached files.


moose-code (commented):
Almost all makes sense except we still use some of the cache files like token-1.json etc, which has token name, decimals and symbols cached for faster resyncing. Will just need to exclude these deletions before merging

stas:
> Almost all makes sense except we still use some of the cache files like token-1.json etc, which has token name, decimals and symbols cached for faster resyncing. Will just need to exclude these deletions before merging

Ty ty. Will take a closer look again.

stas:
Superseded by #325

PR #330, opened 2025-05-22
Change Swell to suspected price connector

moose-code:
This needs investigation, but I think Swell is using the V2 price connector. 
When using V2 price connector, we are not getting any RPC issues. 

When using the V3 price connector, we get the call to getManyRatesWithCustomConnectors reverted.


src/Constants.ts
```diff
@@ -449,20 +449,23 @@ const SWELL_CONSTANTS: chainConstants = {
       return PriceOracleType.V3;
     },
     getAddress: (priceOracleType: PriceOracleType) => {
-      return "0xE50621a0527A43534D565B67D64be7C79807F269";
+      return "0xe58920a8c684CD3d6dCaC2a41b12998e4CB17EfE";
```
stas, line 452: Confirmed, this looks good.

stas (approved):

PR #340, opened 2025-09-15
Fix CLPool Mint tests + decouple mockDB setup from internal logic

FonsSicca:
Implements:
- Fix of `CLPool.Mint` event handler tests
- Decouple logic of writing to DB and "business logic, allowing for a more straigthforward testing without the boilerplate code of constantly setting up mockDB


src/EventHandlers/CLPool/CLPoolMintLogic.ts
```diff
@@ -0,0 +1,115 @@
...
+  | {
+      _type: "TokenNotFoundError";
+      message: string;
+    }
+  | {
+      _type: "LiquidityPoolAggregatorNotFoundError";
+      message: string;
+    };
+
+export function processCLPoolMint(
```
FonsSicca, line 50: main new function that actually only handles "business" logic without using any kind of DB writing

moose-code:
Just wanted to flag we should also update to envio v2.29.0 - lots of improvements. And use the effects api. Will give more thoughts when I'm back in office tomorrow

test/EventHandlers/CLPool/CLPoolMintLogic.test.ts
```diff
@@ -0,0 +1,249 @@
...
+          hash: "0x4444444444444444444444444444444444444444444444444444444444444444",
+        },
+        block: {
+          timestamp: 1000000,
+          number: 123456,
+          hash: "0x5555555555555555555555555555555555555555555555555555555555555555",
+        },
+        chainId: 10,
+        logIndex: 1,
+      };
```
WooSungD, line 34: if the `mockEvents` used in this file are same - lets declare it once and reuse to reduce lines of code 👍

package.json
FonsSicca, line 1: ReScript compilation error that threw a lot of errors across different tests

Might be an envio compilation problem?

More details on `scripts/fix-table-indices.js`

test/EventHandlers/CLPool/CLPool.test.ts
FonsSicca, line 1: Now there is no need for changing anything here, if we maintain that ReScript compilation fix above

scripts/fix-table-indices.js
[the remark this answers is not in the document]
WooSungD: I see, thanks for the detailed explanation and the script for fix.

yo @moose-code, do you know if there is something we can change in velodrome indexing logic to avoid this error (and to prevent having to run script for fix)?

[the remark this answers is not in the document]
DZakh: I'm a little bit out of context here, but could you check that you have the issue with the latest version and if you do, I'll appreciate a github issue.

PR #344, opened 2025-09-17
`CLPoolSwap` refactoring + fixing remaining tests

FonsSicca:
Implements:
- Refactored logic for `CLPoolSwap` event handler
- Fix of remaining tests. **All tests pass now** and only 2 are in "pending". The ones that are in pending are expected because they are within `xit` context, which simply skips the tests. Will analyse if it makes sense to not skipped it in future PRs.

### DIFF
--- a/src/EventHandlers/CLPool.ts
+++ b/src/EventHandlers/CLPool.ts
@@ -19,10 +19,10 @@ import type {
 } from "generated";
 import { updateLiquidityPoolAggregator } from "../Aggregators/LiquidityPoolAggregator";
 import { normalizeTokenAmountTo1e18 } from "../Helpers";
-import { abs, multiplyBase1e18 } from "../Maths";
+import { multiplyBase1e18 } from "../Maths";
 import { fetchPoolLoaderData } from "../Pools/common";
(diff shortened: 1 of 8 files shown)


src/EventHandlers/CLPool.ts
FonsSicca, line 1: decoupled business logic from DB logic

test/EventHandlers/CLPool/CLPool.test.ts
FonsSicca, line 1: tests that were here still exist (e.g. mockDB setup, etc)

Just had to fix some of the parameters (`chainID` 1 is ethereum and it's not specified in `config.yaml` for example)

test/PriceOracle.test.ts
```diff
@@ -312,6 +318,7 @@ describe("PriceOracle", () => {
               test.tokenAddress,
               blockNumber,
               chainId,
+              5000000n, // with 1 million the simulationContract reverts
```
FonsSicca, line 321: this was reverting without the `gasLimit` parameter, that is why I needed to add

WooSungD (approved):

PR #351, opened 2025-09-25
`PoolFees` event handler refactoring and user centric entity creation

FonsSicca:
Implements:
- Similar refactoring pattern to `PoolFees` event handler
- Started transition towards a user centric entity creation. Will update the schema and the codebase along the way


src/Aggregators/Users.ts
FonsSicca, line 1: module that centralises functions related to `User` entity

PR #356, opened 2025-09-29
Envio Effects API refactor for RPC calls

### LINKED ISSUE #353: Use Envio's Effect API
See where it's most appropriate to use this feature, as probably there are RPC calls in the indexer that can be substituted by this and improvement overall readibility and performance

https://docs.envio.dev/docs/HyperIndex/effect-api

FonsSicca:
closes #353

Implements:
- Envio's Effects API feature for RPC calls

### DIFF
--- a/src/Effects/Index.ts
+++ b/src/Effects/Index.ts
@@ -3,8 +3,6 @@ export {
   getTokenDetails,
   getTokenPrice,
   getTokenPriceData,
-  fetchTokenDetails,
-  fetchTokenPrice,
 } from "./Token";
 
 // Dynamic fee-related effects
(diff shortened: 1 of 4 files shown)


src/Effects/Index.ts
WooSungD, line 1: surprised by the number of external calls the indexer is making..

src/Effects/Index.ts
```diff
@@ -0,0 +1,26 @@
+// Token-related effects
+export {
+  getTokenDetails,
+  getTokenPrice,
+  getTokenPriceData,
+  fetchTokenDetails,
+  fetchTokenPrice,
```
WooSungD, line 7: I see we use the effect function (i.e. ones starting with `get`) in the indexer logic and the `fetch` functions only seem to be used for testing.

Is the separate testing of `fetch` functions necessary here?

src/Effects/Index.ts
moose-code: Agreed. I think they can hopefully be cut down, I'm not sure what is necessary vs not.

PR #358, opened 2025-10-03
Pool Burn and Mint refactor + user and pool entity refactor

FonsSicca:
Implements:
- Pool mint and burn refactoring
- Refactoring of the way user entity is created


schema.graphql
```diff
@@ -75,18 +75,30 @@ type LiquidityPoolAggregatorSnapshot {
   timestamp: Timestamp! # timestamp of last update
 }
 
-# Entity for tracking a unique address in the `to` or `from` of a swap event
-# Other liquidity pools don't count as users
-type User {
-  id: ID! # user address
+# Entity for tracking user activity and positions in specific pools
+type UserStatsPerPool {
```
FonsSicca, line 79: an user can have activity on multiple pools, so it's best to have an entity per ID where ID = `userAddress_poolAddress_chainId`

src/Aggregators/LiquidityPoolAggregator.ts
```diff
@@ -153,3 +154,52 @@ export async function updateLiquidityPoolAggregator(
     setLiquidityPoolAggregatorSnapshot(updated, timestamp, context);
   }
 }
+
+/**
```
FonsSicca, line 158: grouped here pool loading logic

src/EventHandlers/Pool/PoolBurnAndMintLogic.ts
FonsSicca, line 1: module that is used in both pool mint and pool burn event handlers, because they basically share the same logic

schema.graphql
FonsSicca: previously there the User entity didn't have a pool-specific field. So basically all the stats would be from all pools  on aggregate

I think separating and having a user entity per pool per chain allows to have more granularity that i'm almost certain that will be useful for metrics

PR #360, opened 2025-10-03
Pool Swap and Sync entity refactor

FonsSicca:
Implements:
- Specific entity refactor to Pool Swap and Sync events


schema.graphql
FonsSicca, line 1: no longer needed

WooSungD (approved):
nice!

PR #370, opened 2025-10-20
Optimize `Pool` event handlers, fix `PoolSync` logic, and improve user stats handling

FonsSicca:
Implements:

- Remove `loaderReturn` compatibility from `Pool.Fees`, `Pool.Swap`, and `Pool.Sync` handlers for cleaner code

- Optimize Pool handlers by pre-computing checksummed addresses at the beginning of each handler

- Remove error field from PoolFeesResult, PoolSwapResult, and PoolSyncResult interfaces as error handling is now centralized

- Fix `PoolSync` event logic to calculate incremental changes from total reserves while maintaining cumulative fields in aggregator (i.e. pool entity)

- Add `calculateTotalLiquidityUSD` helper function to compute total liquidity from current reserves with proper token normalization

- Improve `PoolSyncLogic`to handle edge cases: no tokens (keep existing values), zero amounts (set to zero), and normal cases (incremental changes)

- Simplify `UserStatsPerPool` update function to use direct field access pattern instead of complex field extraction

- Update `Pool.ts` to map `UserLiquidityDiff` fields directly to `UserStatsPerPool`fields (e.g. `netLiquidityAddedUSD` → `currentLiquidityUSD`, `amount0` → `currentLiquidityToken0`, `amount1` → `currentLiquidityToken1`)

- Add `numberOfSwaps` field to `UserSwapDiff` interface and update `processPoolSwap` to include swap count

- Fix all Pool-related test files to match updated function signatures and interface changes

- Update test expectations to match actual interface properties (e.g. remove non-existent fields like `userAddress`, `chainId`, `timestamp`)

### DIFF
--- a/src/EventHandlers/Pool/PoolFeesLogic.ts
+++ b/src/EventHandlers/Pool/PoolFeesLogic.ts
@@ -7,94 +7,52 @@ import type {
 import { updateFeeTokenData } from "../../Helpers";
 
 export interface UserDiff {
-  userAddress: string;
-  chainId: number;
-  feesContributedUSD: bigint;
-  feesContributed0: bigint;
-  feesContributed1: bigint;
-  timestamp: Date;
+  totalFeesContributedUSD: bigint;
+  totalFeesContributed0: bigint;
+  totalFeesContributed1: bigint;
+  lastActivityTimestamp: Date;
 }
 
 export interface PoolFeesResult {
   liquidityPoolDiff?: Partial<LiquidityPoolAggregator>;
   userDiff?: UserDiff;
-  error?: string;
 }
-
-export type PoolFeesLoaderReturn =
-  | {
-      _type: "success";
-      liquidityPoolAggregator: LiquidityPoolAggregator;
-      token0Instance: Token | undefined;
-      token1Instance: Token | undefined;
-    }
-  | {
-      _type: "TokenNotFoundError";
-      message: string;
-    }
-  | {
-      _type: "LiquidityPoolAggregatorNotFoundError";
-      message: string;
-    };
 
 export async function processPoolFees(
   event: Pool_Fees_event,
-  loaderReturn: PoolFeesLoaderReturn,
+  token0Instance: Token | undefined,
+  token1Instance: Token | undefined,
   context: handlerContext,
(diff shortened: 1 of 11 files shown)


src/EventHandlers/Pool/PoolFeesLogic.ts
```diff
@@ -7,94 +7,52 @@ import type {
...
-
 export async function processPoolFees(
   event: Pool_Fees_event,
-  loaderReturn: PoolFeesLoaderReturn,
+  token0Instance: Token | undefined,
+  token1Instance: Token | undefined,
   context: handlerContext,
 ): Promise<PoolFeesResult> {
-  // Handle different loader return types
-  switch (loaderReturn._type) {
```
FonsSicca: again, this type of `loaderReturn` logic was part of the first refactor to maintain a certain degree of backwards compatibility.


Fully removed now

src/EventHandlers/Pool/PoolSyncLogic.ts
```diff
@@ -4,81 +4,75 @@ import type {
...
-          liquidityPoolAggregator.totalLiquidityUSD,
-        token0Price:
-          reserveData.token0?.pricePerUSDNew ??
-          liquidityPoolAggregator.token0Price,
-        token1Price:
-          reserveData.token1?.pricePerUSDNew ??
-          liquidityPoolAggregator.token1Price,
-        lastUpdatedTimestamp: new Date(event.block.timestamp * 1000),
-      };
+    // Calculate total liquidity USD from the new total reserves
```
FonsSicca: Mint, Burn and Sync events affect `reserve0`, `reserve1` and `totalLiquidityUSD` on the pool entity

amounts from mint and burn are "incremental" amounts, meaning if we want to update `totalLiquidityUSD`, we must add/subtract those event-derived amount

the parameters from the sync event are cumulative amounts directly, meaning that if we want to update `totalLiquidityUSD`, we would need to substitute directly and not add or subtract

To have compatibility between these events and not increase number of fields, i simply maintained `reserve0`, `reserve1` and `totalLiquidityUSD` as cumulative quantities when updating pool entity and calculated the incremental changes here in the `Sync` event

PR #382, opened 2025-10-28
Fetching global fee (base + dynamic) for each pool + `CustomFeeSwapModule` for superchain chains

FonsSicca:
Implements:
- RPC contract call to fetch current global fee for pool
- `CustomFeeSwapModule` which is basically almost identical to `DynamicFeeSwapModule` - however, the latter is only available on OP and Base

### DIFF
--- a/src/Aggregators/LiquidityPoolAggregator.ts
+++ b/src/Aggregators/LiquidityPoolAggregator.ts
@@ -38,7 +38,7 @@ export async function updateDynamicFeePools(
 
   if (!dynamicFeeGlobalConfigs || dynamicFeeGlobalConfigs.length === 0) {
     context.log.error(
-      `No dynamic fee global config found for chain ${chainId}`,
+      `No dynamic fee global config found for chain ${chainId}. No update to currentFee will be performed.`,
     );
     return;
   }
--- a/src/Effects/DynamicFee.ts
+++ b/src/Effects/DynamicFee.ts
@@ -20,11 +20,10 @@ export async function fetchCurrentFee(
       `[fetchCurrentFee] Fetching current fee for pool ${poolAddress} on chain ${chainId} at block ${blockNumber}`,
     );
     const DynamicFeePoolABI = require("../../abis/DynamicFeeSwapModule.json");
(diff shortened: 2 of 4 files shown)


src/Aggregators/LiquidityPoolAggregator.ts
```diff
@@ -13,8 +8,6 @@ import type {
 
 const UPDATE_INTERVAL = 60 * 60 * 1000; // 1 hour
 
-const DYNAMIC_FEE_START_BLOCK = 131341414; // Starting from this block to track dynamic fee pools
```
FonsSicca: didn't understand why this was here. Either way, previous iteration of this logic had an `if` which excluded every chain except for OP

src/Effects/DynamicFee.ts
```diff
@@ -20,11 +20,10 @@ export async function fetchCurrentFee(
       `[fetchCurrentFee] Fetching current fee for pool ${poolAddress} on chain ${chainId} at block ${blockNumber}`,
     );
     const DynamicFeePoolABI = require("../../abis/DynamicFeeSwapModule.json");
-    const CustomFeeSwapModuleABI = require("../../abis/CustomFeeSwapModule.json");
 
     const { result } = await ethClient.simulateContract({
       address: dynamicFeeModuleAddress as `0x${string}`,
-      abi: CustomFeeSwapModuleABI,
+      abi: DynamicFeePoolABI,
```
FonsSicca: for the `getFee` function the ABI's are basically the same, so no need to overcomplicate this here imo

src/EventHandlers/CLPool.ts
FonsSicca: fixing some updates regarding User entity

src/Effects/DynamicFee.ts
```diff
@@ -3,58 +3,13 @@ import type { logger as Envio_logger } from "envio/src/Envio.gen";
 import type { PublicClient } from "viem";
 import { CHAIN_CONSTANTS } from "../Constants";
 
-/**
```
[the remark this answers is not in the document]
WooSungD: strange we were making these SC calls + indexing the values

PR #409, opened 2025-11-11
Miscellaneous changes

FonsSicca:
Implements:
- More retries (with higher delay) for rate limit error, when fetching token price)
- Removing redundant code
- Updating token price in `Voter.ts` and `SuperChainLeafVoter.ts`
- Changes to updates made to pool entity in `GaugeKilled` event handler
- More tests to increase test coverage

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -231,7 +231,7 @@ networks:
         address:
           - 0xd4C1905BB1D26BC93DAC913e13CaCC278CdCC80D
   - id: 8453 # Base
-    start_block: 36953918
+    start_block: 0
     contracts:
       - name: DynamicFeeSwapModule
         address:


src/Effects/Token.ts
FonsSicca: more retries, with even higher delay for the occasional case where the previous retries don't work

src/EventHandlers/CLFactory.ts
```diff
@@ -31,12 +30,6 @@ CLFactory.PoolCreated.handler(async ({ event, context }) => {
     context,
   );
 
-  // Apply liquidity pool aggregator updates
```
FonsSicca: this was redudantm we simply set directly the pool entity (this is the `PoolCreated` event handler)

src/EventHandlers/Voter/SuperchainLeafVoter.ts
```diff
@@ -161,8 +161,19 @@ SuperchainLeafVoter.DistributeReward.handler(async ({ event, context }) => {
     return;
   }
 
-  const result = await computeVoterDistributeValues({
+  // Refresh reward token price if it's zero (token was just created or price fetch failed previously)
```
FonsSicca: `computeVoterDistributeValue` computed USD value so we need updated token price data

before we added this, we didn't update so there were cases, specially in the beginning of the indexer sync where the prices were 0.

So now we update them here

test/Aggregators/OUSDTSwaps.test.ts
FonsSicca: from here downwards is a bunch of tests that are created to tests the missing modules. This also increased the project's test coverage

WooSungD:
Nice! Whats the latest test coverage as per this PR?

PR #411, opened 2025-11-12
Adding contracts to `config.yaml` + factory addresses in `Voter` modules

FonsSicca:


config.yaml
FonsSicca: adding missing contracts in some chains

WooSungD (approved):

PR #413, opened 2025-11-13
`Await` for some `async` missing functions + other minor fixes

FonsSicca:
Implements:
- `await` for `async` functions. In some calls, `await` wasn't present
- `rollback_on_reorg` to `true` based on documentation recommendation. https://docs.envio.dev/docs/HyperIndex/config-schema-reference#rollbackonreorg . For developing, we can temporarily set to `false`
- Other minor fixes


src/EventHandlers/CLPool.ts
FonsSicca: added `await`

PR #415, opened 2025-11-14
Removing `isPreload` + token entity updates

FonsSicca:
See individually the commits for better readability

Implements:
- Removal of `isPreload` check of all event handlers. Initially I thought that this would be the default (and therefore, the most accurate to have better performance) but after talking with envio's team it seems like we should start without `isPreload` 
- Added more `Promise.all` for more parallelization and, in principle, better performance
- Implemented missing `Token` entity updates and USD price refresh

### DIFF
--- a/src/Constants.ts
+++ b/src/Constants.ts
@@ -28,6 +28,13 @@ export const SECONDS_IN_AN_HOUR = BigInt(3600);
 export const SECONDS_IN_A_DAY = BigInt(86400);
 export const SECONDS_IN_A_WEEK = BigInt(604800);
 
+// Effect rate limit constants (calls per second)
+export const EFFECT_RATE_LIMITS = {
+  TOKEN_EFFECTS: 4000, // Token details and price fetching effects
+  VOTER_EFFECTS: 4000, // Voter-related effects
+  DYNAMIC_FEE_EFFECTS: 4000, // Dynamic fee effects
+} as const;
+
 export const OUSDT_ADDRESS = "0x1217BfE6c773EEC6cc4A38b5Dc45B92292B6E189";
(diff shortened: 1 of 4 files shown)


src/EventHandlers/VotingReward/VotingRewardSharedLogic.ts
```diff
@@ -107,21 +108,64 @@ export async function processVotingRewardClaimRewards(
...
-  const rewardTokenData = await context.effect(getTokenPriceData, {
-    tokenAddress: data.reward,
-    blockNumber: data.blockNumber,
-    chainId: data.chainId,
-  });
+  // Get reward token and refresh price (refreshTokenPrice handles the update internally)
+  const rewardTokenId = TokenIdByChain(data.reward, data.chainId);
+  let rewardToken = await context.Token.get(rewardTokenId);
+
+  if (!rewardToken) {
```
FonsSicca: if the token entity doesn't exist, we create it

src/EventHandlers/VotingReward/VotingRewardSharedLogic.ts
```diff
@@ -107,21 +108,64 @@ export async function processVotingRewardClaimRewards(
...
+      pricePerUSDNew: rewardTokenPriceData.pricePerUSDNew,
+      lastUpdatedTimestamp: new Date(data.timestamp * 1000),
+      isWhitelisted: true,
+    };
+
+    context.Token.set(newToken);
+    rewardToken = newToken;
+  }
+
+  const updatedRewardToken = await refreshTokenPrice(
```
FonsSicca: we then proceed to update price

PR #420, opened 2025-11-24
Definitive fix for non fungible position entities

FonsSicca:
Implements:

- Fixes critical bug in NFPM token lookup: corrects Token ID format from `${chainId}_${address}` to `${address}-${chainId}` (matching TokenIdByChain), resolving `amountUSD` returning 0 when tokens exist in database

- Ensures `cleanupOrphanedPlaceholders` correctly deletes placeholder entities (`tokenId = 0n`) after they're updated with actual `tokenId`, preventing orphaned placeholder positions from accumulating in the database

- Enhances test assertions in Helpers.test.ts and NFPMLogic.test.ts with exact equality checks, calculating expected values using the same functions as implementation for better test reliability

- Replaces simplified tick math approximations in Helpers.test.ts with actual TickMath.getSqrtRatioAtTick() calls for accurate position amount calculations

### DIFF
--- a/src/EventHandlers/CLPool.ts
+++ b/src/EventHandlers/CLPool.ts
@@ -326,7 +326,7 @@ CLPool.Mint.handler(async ({ event, context }) => {
   // Format: ${chainId}_${fullTxHash}_${logIndex} (without 0x prefix)
   const id = `${event.chainId}_${event.transaction.hash.slice(2)}_${event.logIndex}`;
   context.NonFungiblePosition.set({
-    id: placeholderId, // Placeholder ID - permanent, never changes
+    id: id, // permanent, never changes
     chainId: event.chainId,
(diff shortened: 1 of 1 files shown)


src/Effects/Token.ts
```diff
@@ -222,6 +223,153 @@ export async function fetchTokenPrice(
   };
 }
 
+/**
+ * Core logic for fetching sqrtPriceX96 from pool's slot0 function
+ * Includes fallback to public RPC when private RPC fails with historical state errors
+ */
+export async function fetchSqrtPriceX96(
```
FonsSicca: new function to fetch `SqrtPriceX96`

This is needed because in between 2 consecutive `NFPM` events, the pool price can change so we need to recalculate the actual entities amount so that, afterwards, the summing (if the event is `IncreaseLiquidity`) or the subtraction (if `DecreaseLiquidity`) are right

WooSungD (approved):

PR #421, opened 2025-11-24
New rpc logging

FonsSicca:


src/Constants.ts
```diff
@@ -85,6 +85,12 @@ export enum PriceOracleType {
   V1 = "v1",
 }
 
+/**
```
FonsSicca: limit for rpc call to timeout

src/Effects/Token.ts
```diff
@@ -120,6 +123,21 @@ export async function fetchTokenPrice(
           blockNumber: BigInt(blockNumber),
           gas: currentGasLimit,
         });
+        const attemptDuration = Date.now() - attemptStartTime;
```
FonsSicca: we now log the request that are slow and very slow, if further debugging is needed

PR #430, opened 2025-12-01
Further work on prices

FonsSicca:
Implements:
- Moralis API (https://docs.moralis.com/web3-data-api/evm) as fallback for when `historical state is not available` keeps getting thrown. Tried on some of the affected tokens for the mentioned blocks and it seems to be working. Two problems though: it isn't probably fully robust for all tokens in all necessary blocks and we need to get a paid plan if we go ahead with this
- Slight changes to `getSqrtPriceX96` effect for better cache optimisation and to avoid indexer crashing


scripts/fetch_prices_v3.ts
FonsSicca: just a script to test price fetching using oracle v3

src/PriceOracle.ts
```diff
@@ -104,7 +108,66 @@ export async function refreshTokenPrice(
         gasLimit,
       }),
     ]);
-    const currentPrice = priceData.pricePerUSDNew;
+    let currentPrice = priceData.pricePerUSDNew;
+
+    // Try fetching price from Moralis API as fallback when oracle returns 0
```
FonsSicca: new Moralis API fallback. It works for many tokens where historial state is not available on the RPC calls, but: 1) it's probable that it isn't totally robust to all tokens in all blocks needed; 2) we need a paid plan which might be expensive

src/PriceOracle.ts
```diff
@@ -164,17 +227,27 @@ export async function refreshTokenPrice(
...
-    // Create new TokenPrice entity
-    const tokenPrice: TokenPriceSnapshot = {
-      id: TokenIdByBlock(token.address, chainId, blockNumber),
-      address: toChecksumAddress(token.address),
-      pricePerUSDNew: currentPrice,
-      chainId: chainId,
-      isWhitelisted: token.isWhitelisted,
-      lastUpdatedTimestamp: new Date(blockTimestampMs),
-    };
+    if (
```
FonsSicca: avoiding duplicated `TokenPriceSnapshot` in the same block

PR #435, opened 2025-12-05
Optimising `getSqrtX96` effect and avoiding indexer crashing

FonsSicca:


src/Effects/Token.ts
FonsSicca: cherry picked from the other PR: decided to isolate this

PR #439, opened 2025-12-09
Further fixes to ALM Wrapper entity

FonsSicca:
Implements:
- New `TotalSupplyLimited` event for initial track of total LP tokens
- Logic correction of `amount0` and `amount1` in ALM LP Wrapper entity: 
     - Considers initial seeding of the LP position with `amount0` and `amount1` by filtering `NonFungiblePosition` entity created in the same transaction hash (and with other equal fields)
     - Fix to `Deposit` and `Withdrawal` event handler logic: now accounts for the actual amounts deposits/withdrawals (besides recalculating the amounts based on liquidity and `sqrtPriceX96` which it was the only thing that was doing previously)

### DIFF
--- a/src/EventHandlers/ALM/DeployFactory.ts
+++ b/src/EventHandlers/ALM/DeployFactory.ts
@@ -1,4 +1,5 @@
 import { ALMDeployFactory } from "generated";
+import { ALM_TotalSupplyLimitUpdated_event } from "generated/src/db/Entities.res";
 import { toChecksumAddress } from "../../Constants";
 
 ALMDeployFactory.StrategyCreated.contractRegister(({ event, context }) => {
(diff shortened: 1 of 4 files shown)


src/EventHandlers/ALM/DeployFactory.ts
```diff
@@ -31,14 +32,53 @@ ALMDeployFactory.StrategyCreated.handler(async ({ event, context }) => {
     await context.NonFungiblePosition.getWhere.mintTransactionHash.eq(
       event.transaction.hash,
     );
-  if (!nonFungiblePositions || nonFungiblePositions.length === 0) {
+  const matchingNonFungiblePositions =
+    nonFungiblePositions?.filter(
```
FonsSicca: for a robust matching, we do not only match by transaction hash but we make sure other properties are also the same because there can be more than 1 non fungible position created in the same tx

src/EventHandlers/ALM/DeployFactory.ts
```diff
@@ -48,12 +88,12 @@ ALMDeployFactory.StrategyCreated.handler(async ({ event, context }) => {
     pool: toChecksumAddress(pool),
     token0: toChecksumAddress(token0),
     token1: toChecksumAddress(token1),
-    // Wrapper-level aggregations (updated by Deposit/Withdraw events)
-    amount0: 0n,
-    amount1: 0n,
-    lpAmount: 0n,
+
+    amount0: amount0,
```
FonsSicca: now we correctly initialise the `ALM_LP_Wrapper` with correct values of `amount0`, `amount1` and `lpAmount`

src/EventHandlers/ALM/DeployFactory.ts
```diff
@@ -31,14 +32,53 @@ ALMDeployFactory.StrategyCreated.handler(async ({ event, context }) => {
...
+  }
+
+  // there should, in principle, one unique non fungible position that has
+  // simultaneously the same transaction hash,tickLower, tickUpper, liquidity and token0 and token1
+  const tokenId = matchingNonFungiblePositions[0].tokenId;
+  const amount0 = matchingNonFungiblePositions[0].amount0;
+  const amount1 = matchingNonFungiblePositions[0].amount1;
+
+  // Fetching LP tokens supply from TotalSupplyLimitUpdated event
+  const ALM_TotalSupplyLimitUpdated_event =
```
FonsSicca: the new entity is used here, to later fetch the total LP tokens that were created when initialising the LP wrapper

src/EventHandlers/ALM/LPWrapper.ts
```diff
@@ -217,3 +213,25 @@ ALMLPWrapper.Transfer.handler(async ({ event, context }) => {
     updateUserStatsPerPool(UserStatsToDiff, userStatsTo, timestamp, context),
   ]);
 });
+
+/**
```
FonsSicca: new event handler

WooSungD (approved):

PR #443, opened 2025-12-12
[1] Gauge logic refactoring

FonsSicca:
Implements:
- Major gauge logic refactoring: not only performance improvements (e.g. parallelization through `Promise.all`) but also standardizing code patterns as well as additional/fixing logic and additional fields in schema


schema.graphql
```diff
@@ -59,6 +59,7 @@ type LiquidityPoolAggregator {
   numberOfGaugeRewardClaims: BigInt! @config(precision: 76) # number of gauge reward claims
   totalGaugeRewardsClaimedUSD: BigInt! @config(precision: 76) # total gauge rewards claimed in USD
   totalGaugeRewardsClaimed: BigInt! @config(precision: 76) # total gauge rewards claimed in token units
+  currentLiquidityStaked: BigInt! @config(precision: 76) # current liquidity staked in gauge in token units
```
FonsSicca: added gauge token units field

src/Helpers.ts
```diff
@@ -316,6 +321,119 @@ export function calculatePositionAmountsFromLiquidity(
   };
 }
 
+/**
```
FonsSicca: new function that calculates the USD value of staked liquidity

For CL pools we calculate the amounts0 and 1 from the liquidity value and then calculate USD value. In this case, `amount` is liquidity in the traditional definition of uniswap

For non-CL pools, we call the new effect to get the total  supply of the LP token to then derive the proportional amounts of token0 and token1 that the number of LP tokens represent. Then we calculate the respective USD value. In this case, `amount` is the number of LP tokens and not liquidity in the traditional definition of uniswap

PR #447, opened 2025-12-15
`SuperChainLeafVoter` event handler deletion & general ABI json for all events

FonsSicca:
Implements:
- Deletion of events that were wrongly set for `SuperChainLeafVoter` - this was due to the assumption that its ABI would be equal to `Voter.ts` which is not. Therefore, the way veNFT deposits/withdrawals are tracked depends whether `Voter.ts` (for optimism and base) or `SuperChainLeafVoter.ts` is the one handling such operation, since their events are slightly different. More on this in later PRs
- Standardizing events indexed through the same ABI file path pattern

## Summary by CodeRabbit

* **New Features**
  * Added ABI/interface definitions for CLFactory, Pool, PoolFactory, SuperchainLeafVoter, and VeNFT.

* **Refactor**
  * Normalized event schemas to generic event names in config.
  * Removed Voted and DistributeReward event handlers from SuperchainLeafVoter.

* **Tests**
  * Updated tests to use WhitelistToken flows and adjusted mock field naming for VeNFT.

* **Chores**
  * Added repo tooling config (.coderabbit.yaml) and ABI references.

### DIFF
--- a/test/EventHandlers/Voter/SuperchainLeafVoter.test.ts
+++ b/test/EventHandlers/Voter/SuperchainLeafVoter.test.ts
@@ -1,37 +1,10 @@
 import { expect } from "chai";
 import { MockDb, SuperchainLeafVoter } from "generated/src/TestHelpers.gen";
-import type {
-  LiquidityPoolAggregator,
-  Token,
-  UserStatsPerPool,
-} from "generated/src/Types.gen";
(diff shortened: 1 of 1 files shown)


WooSungD:
What is the reason for pointing to all the ABI files from the config in this commit?

PR #450, opened 2025-12-17
Voting logic for leaf Pools

FonsSicca:
Implements:

- Support for correct tracking of voting of leaf pools (CL and non-CL)

## Summary by CodeRabbit

* **New Features**
  * Added support for RootCLPool contracts with automatic cross-chain resolution to corresponding leaf pools
  * Enhanced pool data tracking with cross-chain identification fields and tick spacing information

* **Chores**
  * Added RootCLPoolFactory contract integration and configuration across supported networks

### DIFF
--- a/src/Aggregators/LiquidityPoolAggregator.ts
+++ b/src/Aggregators/LiquidityPoolAggregator.ts
@@ -619,6 +619,6 @@ export function createLiquidityPoolAggregatorEntity(params: {
     feeCap: undefined,
     scalingFactor: undefined,
     currentFee: undefined,
-    rootPoolMatchingHash: `${chainId}_${token0Address}_${token1Address}_${tickSpacing}`,
+    rootPoolMatchingHash: `${chainId}_${token0Address}_${token1Address}_${(tickSpacing ? BigInt(tickSpacing) : 0n).toString()}`,
   };
 }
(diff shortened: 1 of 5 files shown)


src/EventHandlers/Voter/Voter.ts
```diff
@@ -123,7 +130,13 @@ Voter.Abstained.handler(async ({ event, context }) => {
   // Load pool data and user data concurrently for better performance
   const poolAddress = toChecksumAddress(event.params.pool);
   const [poolData, userData] = await Promise.all([
-    loadPoolData(poolAddress, event.chainId, context),
+    loadPoolDataOrRootCLPool(
```
FonsSicca: this is the main change

so basically both `Voted` and `Abstained` events can have as `pool` parameter a root pool (which isn't a real pool)

previously we would try to access a pool entity for a root pool address -> error occured because since root pool isn't a real pool, it isn't created by `PoolCreated` which actually creates a pool entity (root pools are created by `RootPoolCreated` or `RootCLPoolCreated`)

now with `loadPoolDataOrRootCLPool` we first try to fetch a pool entity from the `pool` parameter of the event. If the pool entity doesn't exist, it might be a root pool. This is where the mapping enters into action. We try to fetch the leaf pool (which is the real pool) entity through the mapping we've made between root <-> pool

WooSungD (approved):

PR #451, opened 2025-12-18
ALM - Adding old deploy factory and core contracts

FonsSicca:
Implements:
- New ALM deploy factory contract (and respective event handler) and new ALM core contract. Deploy factory contract has `StrategyCreated` event with a different structure to the one already existing so a new event handler is needed. This isn't the case for ALM core, where both the old and new contracts have the same ABI for the relevant events, therefore can use the same event handler

## Summary by CodeRabbit

* **New Features**
  * Added V1 and V2 variants for ALM Deploy Factory and ALM LP Wrapper, plus network mappings for both versions
  * New event handlers for strategy creation and for LP deposit/withdraw/transfer flows
  * User statistics now track ALM address

* **Chores**
  * Configuration updated to reference versioned ALM contracts across networks

* **Tests**
  * Comprehensive test suites added for DeployFactory and LPWrapper V1/V2 handlers

### DIFF
--- a/src/EventHandlers/ALM/LPWrapperV2.ts
+++ b/src/EventHandlers/ALM/LPWrapperV2.ts
@@ -222,8 +222,7 @@ ALMLPWrapperV2.Transfer.handler(async ({ event, context }) => {
  * (e.g., StrategyCreated) can seed `lpAmount` from the latest supply.
  */
 ALMLPWrapperV2.TotalSupplyLimitUpdated.handler(async ({ event, context }) => {
-  const { newTotalSupplyLimit, totalSupplyLimitOld, totalSupplyCurrent } =
-    event.params;
+  const { totalSupplyCurrent } = event.params;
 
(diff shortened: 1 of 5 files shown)


src/EventHandlers/ALM/DeployFactoryV1.ts
FonsSicca: new handler for the old deploy factory contract

src/EventHandlers/ALM/LPWrapperV2.ts
FonsSicca: renamed

src/EventHandlers/ALM/LPWrapperV1.ts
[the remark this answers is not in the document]
FonsSicca: V1 is always the old one and V2 is always the new one

we always had the new one. It was missing the old one

Afaik this is only necessary on base and optimism

PR #467, opened 2025-12-29
[1] `Diff` variables keys renaming

FonsSicca:
Main change is relative to a more readable and intuitive interpretation of `diff`-like variable fields which are incremental (and not cumulative): for these, `incremental` prefix was added

## Summary by CodeRabbit

* **Refactor**
  * Switched many aggregators and event flows from cumulative totals to incremental diff semantics for pools, users, veNFTs and LP wrappers.

* **Behavior**
  * Non-fungible positions now use absolute amounts and propagate explicit last-updated timestamps; some aggregated totals are computed from incremental deltas.

* **Event Handling**
  * Event payloads and handler outputs now emit incremental-prefixed metrics (e.g., incrementalTotal..., incrementalCurrent...) for per-event deltas.

* **Tests**
  * Test suites updated to assert new incremental field shapes and adjusted return/result layouts.

### DIFF
--- a/src/Aggregators/LiquidityPoolAggregator.ts
+++ b/src/Aggregators/LiquidityPoolAggregator.ts
@@ -36,7 +36,7 @@ export interface LiquidityPoolAggregatorDiff {
   incrementalNumberOfSwaps: bigint;
   incrementalTotalEmissions: bigint;
   incrementalTotalEmissionsUSD: bigint;
-  totalBribesUSD: bigint;
+  incrementalTotalBribesUSD: bigint;
   incrementalTotalFlashLoanFees0: bigint;
   incrementalTotalFlashLoanFees1: bigint;
   incrementalTotalFlashLoanFeesUSD: bigint;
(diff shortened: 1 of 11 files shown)


src/Aggregators/LiquidityPoolAggregator.ts
```diff
@@ -173,74 +173,84 @@ export async function updateLiquidityPoolAggregator(
   let updated: LiquidityPoolAggregator = {
     ...current,
     // Handle cumulative fields by adding diff values to current values
-    reserve0: (diff.reserve0 ?? 0n) + current.reserve0,
-    reserve1: (diff.reserve1 ?? 0n) + current.reserve1,
+    reserve0: (diff.incrementalReserve0 ?? 0n) + current.reserve0,
```
FonsSicca: the main idea here when updating is that any field of `LiquidityPoolAggregator` which is cumulative should have `incremental` variable coming from `diff` and the "total current cumulative" coming from `current`

for direct substitution variables, the `diff` variable shouldn't have `incremental` as prefix

WooSungD (approved):

PR #478, opened 2026-01-14
Effects refactor

FonsSicca:
Implements:
- General fallback retry to current block if rounded block fails
- Standardizing code pattern across effects modules
- Improving testing coverage of effects

## Release Notes

* **New Features**
  * Added retry logic for failed blockchain data fetches using alternate block numbers to improve reliability.
  * Implemented graceful fallback behavior for dynamic fees and token prices when data retrieval fails.

* **Bug Fixes**
  * Improved error handling to prevent crashes when external blockchain calls fail; system now continues with sensible defaults.
  * Enhanced stability by allowing operations to proceed when optional data sources become unavailable.

### DIFF
--- a/src/Aggregators/LiquidityPoolAggregator.ts
+++ b/src/Aggregators/LiquidityPoolAggregator.ts
@@ -102,26 +102,29 @@ export async function updateDynamicFeePools(
 
   const dynamicFeeModuleAddress = dynamicFeeGlobalConfigs[0].id;
 
-  try {
-    // base fee + dynamic fee
-    const currentFee = await context.effect(getCurrentFee, {
-      poolAddress,
-      dynamicFeeModuleAddress,
-      chainId,
-      blockNumber,
-    });
+  // base fee + dynamic fee
+  const currentFee = await context.effect(getCurrentFee, {
+    poolAddress,
+    dynamicFeeModuleAddress,
+    chainId,
+    blockNumber,
+  });
 
-    // Update the current fee in the pool entity
-    const updated: LiquidityPoolAggregator = {
-      ...liquidityPoolAggregator,
-      currentFee,
-    };
-
-    context.LiquidityPoolAggregator.set(updated);
-  } catch (error) {
-    // No error if the pool is not a dynamic fee pool
+  // If fee is undefined, it means the effect failed - skip update to preserve existing fee
+  if (currentFee === undefined) {
+    context.log.warn(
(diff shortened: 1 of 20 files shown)


src/Aggregators/LiquidityPoolAggregator.ts
```diff
@@ -102,26 +102,29 @@ export async function updateDynamicFeePools(
 
   const dynamicFeeModuleAddress = dynamicFeeGlobalConfigs[0].id;
 
-  try {
```
FonsSicca: no need for `try` `catch` conditions since now error handling as been moved to the actual effect logic (in this case, to `getCurrentFee`

src/Effects/DynamicFee.ts
```diff
@@ -15,31 +17,29 @@ export async function fetchCurrentFee(
   ethClient: PublicClient,
   logger: Envio_logger,
 ): Promise<bigint> {
-  try {
```
FonsSicca: similar explanation as above: `try` and `catch` flow is handled in the actual `getCurrentFee` effect. This `fetchCurrentFee` is the business logic of `getCurrentFee` effect

src/Effects/Token.ts
```diff
@@ -1,56 +1,58 @@
 import { S, createEffect } from "envio";
 import type { logger as Envio_logger } from "envio/src/Envio.gen";
 import type { PublicClient } from "viem";
+import CLPOOL_ABI from "../../abis/CLPool.json";
```
FonsSicca: seems that this is more efficient than using `require` so i made these changes throughout the different effects modules

src/Effects/Token.ts
```diff
@@ -763,25 +738,18 @@ export const getSqrtPriceX96 = createEffect(
...
+        context,
+        "getSqrtPriceX96",
+        {
+          poolAddress,
+          chainId,
+          blockNumber,
+        },
+        undefined,
       );
-      throw readableError;
```
FonsSicca: just a note: there is a change in the way effects return on error. Previously some effects threw an errror, which would likely cause the indexer to crash. 

Now every effects returns `undefined` and the function(s) that call such effect should handle appropriatly and accordingly to this fact

test/Aggregators/LiquidityPoolAggregator.test.ts
FonsSicca: from here downwards is a combination of refactor tests for new approach/code design + including extra tests to improve coverage mainly on `Effects` folder

WooSungD (approved):

PR #480, opened 2026-01-15
Updating correctly `currentFee`for `CLPool` + correct accounting of swap fees

### LINKED ISSUE #473: Implement `updateDynamicFeePools`
Currently `currentFee` is only updated through `updateDynamicFeePools` but this isn't being used right now. One reason is due to the expected high number of calls that could be needed, namely in syncing to latest block

Assess whether it is really necessary to have `currentFee` field on pool entity. At first sight, it seems to be needed, since it is the only way we can compute fees generated through swaps for CLPool (and not collected by unstaked/staked positions!)

For non-CL pools, fees generated can be captured through `Fees` event

FonsSicca:
closes #473 

Implements:
- Actual usage of `updateDynamicFeePools` which makes an RPC call to retrieve current fees for CLPools.
- Accounting of swaps fees in CLPool swaps using above result from RPC call
- Refactor of `CLPoolSwapLogic`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling to stop processing when required fee mapping is missing and surface failures.

* **Refactor**
  * Fee fields are now required to enforce data integrity.
  * Swap calculations refactored for more accurate, granular fee, volume, and liquidity accounting.
  * Dynamic fee update flows now return updated results instead of silent mutations.

* **Tests**
  * Expanded tests to cover dynamic fee behavior, swap helpers, and pool-type branching.

### DIFF
--- a/src/Aggregators/LiquidityPoolAggregator.ts
+++ b/src/Aggregators/LiquidityPoolAggregator.ts
@@ -312,10 +312,15 @@ export async function updateLiquidityPoolAggregator(
     timestamp.getTime() - current.lastSnapshotTimestamp.getTime() >
       UPDATE_INTERVAL
   ) {
-    updated = {
-      ...updated,
-      ...(await updateDynamicFeePools(updated, context, blockNumber)),
-    };
+    // Only update dynamic fees for CL pools (they use dynamic fee modules)
+    // Non-CL pools have their fees fixed at a certain constant level. It can change over time, but we fetch that change
+    // through events.
+    if (updated.isCL) {
+      updated = {
+        ...updated,
+        ...(await updateDynamicFeePools(updated, context, blockNumber)),
+      };
+    }
 
     setLiquidityPoolAggregatorSnapshot(updated, timestamp, context);
   }
--- a/test/Aggregators/LiquidityPoolAggregator.test.ts
+++ b/test/Aggregators/LiquidityPoolAggregator.test.ts
@@ -382,6 +382,13 @@ describe("LiquidityPoolAggregator Functions", () => {
         lastSnapshotTimestamp: oldTimestamp,
       };
 
+      // Mock the effect to track if it's called
+      if (!mockContext.effect) {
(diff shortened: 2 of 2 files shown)


src/Aggregators/LiquidityPoolAggregator.ts
```diff
@@ -312,6 +312,16 @@ export async function updateLiquidityPoolAggregator(
     timestamp.getTime() - current.lastSnapshotTimestamp.getTime() >
       UPDATE_INTERVAL
   ) {
+    // Only update dynamic fees for CL pools (they use dynamic fee modules)
```
FonsSicca: updating here currentFee for CL pools

src/EventHandlers/CLPool/CLPoolSwapLogic.ts
```diff
@@ -6,64 +6,267 @@ import type {
...
+  newReserve0: bigint;
+  newReserve1: bigint;
+  deltaTotalLiquidityUSD: bigint;
+}
+
+/**
+ * Calculates swap volume in USD
+ * @internal Exported for testing purposes only
+ */
+export function _calculateSwapVolume(
```
FonsSicca: using underscore convention for function that are meant to be used within the scope where they are defined (module level)

src/EventHandlers/CLPool/CLPoolSwapLogic.ts
```diff
@@ -6,64 +6,267 @@ import type {
...
   token0Instance: Token | undefined,
   token1Instance: Token | undefined,
   context: handlerContext,
-): Promise<CLPoolSwapResult> {
-  // Process both tokens in parallel using updateTokenData helper
-  // Apply abs() to signed amounts (int256) before processing
-  const swapData = await updateSwapTokenData(
+): SwapFees {
+  // Get the current fee, falling back to baseFee if currentFee is undefined
+  const fee =
```
FonsSicca: at most, fee value is 1h old (which is the frequency at which the currentFee fields is updated)

PR #483, opened 2026-01-19
Remove token holdings from user entity & NFPM refactor

FonsSicca:
Implements:
- Split monolithic `NFPMLogic.ts` into more granular modules
- Removed token holding fields from user entity (CL Pools). This decision was motivated by the fact that token holdings constantly change by any minor price swing of the pool, thus the data for a user would become irrelevant/out-of-date the immediate moment after the update to token holdings was performance (since any price swing would change the balance between holdings in token0 and token1). Thus, the decision is only save the invariant quantity: `liquidity`, `L`, parameter for CL Pools. The correspondent invariant for non-CL pools is `lpBalance` (i.e. number of ERC20 LP tokens), already addressed in a previous PR. 
- Substitution of RPC calls of `sqrtPriceX96` for reads of pool entities. Since `Swap` events have `sqrtPriceX96` data, we have always direct access to updated `sqrtPriceX96` without the need to use RPC calls.

## Summary by CodeRabbit

* **New Features**
  * Added support for Optimism network configuration
  * Enhanced liquidity pool data with real-time price tracking
  * Improved NFT position creation workflow for mint events

* **Refactor**
  * Restructured liquidity accounting from current balances to cumulative added/removed totals
  * Simplified event handling logic for position and liquidity management
  * Updated position tracking to use stable identifiers

### DIFF
--- a/src/EventHandlers/ALM/DeployFactoryV1.ts
+++ b/src/EventHandlers/ALM/DeployFactoryV1.ts
@@ -42,14 +42,14 @@ ALMDeployFactoryV1.StrategyCreated.handler(async ({ event, context }) => {
 
   if (matchingPositions.length === 0) {
     context.log.error(
-      `NonFungiblePosition not found for transaction hash ${event.transaction.hash} matching tickLower ${tickLower}, tickUpper ${tickUpper}, liquidity ${liquidity}. It should have been created by CLPool event handlers.`,
+      `[ALMDeployFactoryV1] NonFungiblePosition not found for transaction hash ${event.transaction.hash} matching tickLower ${tickLower}, tickUpper ${tickUpper}, liquidity ${liquidity}. It should have been created by CLPool event handlers.`,
     );
     return;
   }
 
   if (matchingPositions.length > 1) {
     context.log.warn(
-      `Multiple NonFungiblePositions found for transaction hash ${event.transaction.hash} with the same tick lower ${tickLower}, tick upper ${tickUpper}, liquidity ${liquidity}, token0 ${token0} and token1 ${token1}. Using the first match.`,
+      `[ALMDeployFactoryV1] Multiple NonFungiblePositions found for transaction hash ${event.transaction.hash} with the same tick lower ${tickLower}, tick upper ${tickUpper}, liquidity ${liquidity}, token0 ${token0} and token1 ${token1}. Using the first match.`,
     );
   }
 
@@ -88,7 +88,7 @@ ALMDeployFactoryV1.StrategyCreated.handler(async ({ event, context }) => {
     amount1 = amounts.amount1;
   } else {
     context.log.warn(
-      `Could not fetch sqrtPriceX96 for pool ${pool} to compute amount0/amount1, using 0`,
+      `[ALMDeployFactoryV1] Could not fetch sqrtPriceX96 for pool ${pool} to compute amount0/amount1, using 0`,
     );
   }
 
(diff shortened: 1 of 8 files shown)


config.yaml
FonsSicca: had to add these addresses

This kind of happens from time to time: only through some transaction tracking I can get to these (apparently) non-documented addresses, from the sources I've been visiting

src/EventHandlers/ALM/DeployFactoryV1.ts
```diff
@@ -20,39 +22,75 @@ ALMDeployFactoryV1.StrategyCreated.handler(async ({ event, context }) => {
...
+      tickLower,
+      tickUpper,
+    );
+    amount0 = amounts.amount0;
+    amount1 = amounts.amount1;
+  } else {
+    context.log.warn(
+      `[ALMDeployFactoryV1] Could not fetch sqrtPriceX96 for pool ${pool} to compute amount0/amount1, using 0`,
+    );
+  }
```
FonsSicca: i've stayed with the amount in the ALM entity but i'm open to also remove from it and only add it to a new entity which would be the snapshot version of the ALM entity (as we will to do user and non fungible position entities)

wdyt @WooSungD ?

this basically also means iterating across all ALM wrappers to update their token holdings in a hourly fashion (if we only update/track token holdings in snapshots and if we do hourly intervals)

src/EventHandlers/CLPool/CLPoolSwapLogic.ts
FonsSicca: Swap event had `sqrtPriceX96` and `tick`

This allows to have always the correct `sqrtPriceX96` price and eliminates the need for RPC calls, right? Since swaps are the only way to modify the price

@WooSungD

src/EventHandlers/NFPM/NFPMCommonLogic.ts
FonsSicca: common helper functions for NFPM-related modules

src/EventHandlers/NFPM/NFPMIncreaseLiquidityLogic.ts
```diff
@@ -0,0 +1,121 @@
...
+      context.CLPoolMintEvent.deleteUnsafe(matchingMintEvent.id);
+    }
+  }
+
+  // Calculate increase liquidity diff
+  const nonFungiblePositionDiff = _calculateIncreaseLiquidityDiff(event);
+
+  // Update position with result
+  updateNonFungiblePosition(nonFungiblePositionDiff, position, context);
+}
```
FonsSicca: in-line codes should be sufficiently clear to explain the main logic flow of this

any doubt let me know

src/EventHandlers/NFPM/NFPMLogic.ts
FonsSicca: deleted after the logic separation

src/EventHandlers/NFPM/NFPMTransferLogic.ts
FonsSicca: once again, in-line comments should be sufficiently clear to understand but let me know if any doubts arise. Happy to clarify

src/EventHandlers/NFPM/NFPMTransferLogic.ts
```diff
@@ -0,0 +1,211 @@
...
+ * @param mintEvent - The CLPoolMintEvent to create position from
+ * @param tokenId - The actual token ID from the NFPM.Transfer event
+ * @param owner - The owner address from the NFPM.Transfer event
+ * @param chainId - The chain ID
+ * @param blockTimestamp - The block timestamp
+ * @param context - The handler context for database operations
+ * @returns The newly created stable position entity
+ * @internal
+ */
+export async function _createPositionFromCLPoolMint(
```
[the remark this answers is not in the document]
FonsSicca: is meant to be internally used but need the `export` key to be tested

Given this context, should we still have the `_`?

PR #498, opened 2026-01-21
`sqrtPriceX96` RPC removal + NFPM Increase Liquidity event logic improvement

FonsSicca:
Implements:
- Removal of `sqrtPriceX96` RPC calls. Substituted by reads of pool entities which have updated `sqrtPriceX96` (from `Swap` events)
- Slight refactoring of other handlers, namely `IncreaseLiquidity` in `NFPM.ts` (for a more deterministic outcome)
- Removal of underscore from function names

## Summary by CodeRabbit

* **Bug Fixes**
  * Price is now sourced only from the pool aggregator; missing or zero prices produce warnings and default amounts to zero.

* **Refactor**
  * Internal helpers made public to improve testability.
  * Non-fungible position IDs now include pool address.

* **Behavior Changes**
  * Per-token "current liquidity" diffs removed; token amounts tracked as total added/removed.

* **Tests**
  * Updated test suites to match new price retrieval and identifier semantics; some legacy price-fetch tests removed.

### DIFF
--- a/src/EventHandlers/ALM/DeployFactoryV1.ts
+++ b/src/EventHandlers/ALM/DeployFactoryV1.ts
@@ -65,7 +65,8 @@ ALMDeployFactoryV1.StrategyCreated.handler(async ({ event, context }) => {
 
   let amount0 = 0n;
   let amount1 = 0n;
-  if (sqrtPriceX96 && sqrtPriceX96 !== 0n) {
+  // Treat sqrtPriceX96 === 0n the same as undefined (missing/invalid)
+  if (sqrtPriceX96 !== undefined && sqrtPriceX96 !== 0n) {
     const amounts = calculatePositionAmountsFromLiquidity(
       liquidity,
       sqrtPriceX96,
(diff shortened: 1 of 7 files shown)


src/EventHandlers/ALM/LPWrapperLogic.ts
FonsSicca: removing `sqrtPriceX96` rpc call logic from here

src/Aggregators/UserStatsPerPool.ts
```diff
@@ -3,8 +3,6 @@ import { toChecksumAddress } from "../Constants";
 
 export interface UserStatsPerPoolDiff {
   incrementalCurrentLiquidityUSD: bigint;
-  incrementalCurrentLiquidityToken0: bigint;
-  incrementalCurrentLiquidityToken1: bigint;
```
WooSungD: Just to confirm, I see that the logic for updating these fields were removed in #483, this is just a follow-up change to remove the unused fields entirely right?

PR #540, opened 2026-02-25
Restructuring RPC Calls (Effects)

### LINKED ISSUE #488: Token prices - possible use Alchemy's Prices API instead of on-chain oracle (or mix of both)
Currently we use velodrome's on chain oracle with no fallback if that oracle, for some reason, returns non-expected/incorrect data (e.g. contract doesn't exist at a certain block, 0 price, etc)

One possible mitigation would be to use alchemy's prices API:

https://www.alchemy.com/docs/reference/prices-api-quickstart

A more thorough investigation needs to be put into this to make sure that:
1. It is a reliable  price source 
2. Can make up to missing token prices of on-chain oracle

### LINKED ISSUE #505: Best practices for RPC calls on alchemy
Non priority at the moment but can be improve performance, potentially

Read the following:

https://www.alchemy.com/docs/best-practices-when-using-alchemy

https://www.alchemy.com/docs/how-to-implement-retries

### LINKED ISSUE #508: Remove unnecessary RPC call for gauge liveness in `DistributeReward` handler
## Problem

The `computeVoterDistributeValues` function in `VoterCommonLogic.ts` makes an unnecessary RPC call to `getIsAlive` to check if a gauge is alive. However, this information is already tracked in the database via indexed events and is available in the `LiquidityPoolAggregator` entity.

## Current Behavior

When processing `DistributeReward` events, the code:
1. Loads `currentLiquidityPool` from the database (which contains `gaugeIsAlive`)
2. Makes an RPC call via `getIsAlive` effect to fetch the same information
3. Uses the RPC result (with fallback to `false` if it fails)

This creates unnecessary:
- Network latency
- RPC rate limiting concerns
- Potential points of failure
- Increased indexing costs

## Root Cause Analysis

The `gaugeIsAlive` field in `LiquidityPoolAggregator` is already maintained by event handlers:
- `GaugeCreated` → sets `gaugeIsAlive: true` (Voter.ts:52)
- `GaugeKilled` → sets `gaugeIsAlive: false` (Voter.ts:304)
- `GaugeRevived` → sets `gaugeIsAlive: true` (Voter.ts:329)

Since events are the source of truth and are processed in order, the database value is always accurate and up-to-date.

## Proposed Solution

Remove the `getIsAlive` RPC call and use the `gaugeIsAlive` value from the `currentLiquidityPool` entity that's already loaded in the `DistributeReward` handler.

FonsSicca:
closes #505 
closes #508 
closes #488 

Implements (main changes):
- Centralized RPC via `RpcGateway`
     - New `src/Effects/RpcGateway.ts`: single gateway for all effect RPC (token details, token price, tokens deposited, swap fee, root pool address).
     - Shared retries, timeouts, error handling, and optional batching/coalescing instead of per-effect logic.
     - Effect names use an `EffectType` enum; per-effect rateLimit was removed in favor of gateway-level control.
- Effects split into “fetchers” + thin effect wrappers
     - New `src/Effects/fetchers/ `modules: `Token.ts`, `Voter.ts`, `RootPool.ts`, `SwapFee.ts` hold the actual RPC/contract calls (viem, ABIs).
     - `Token.ts`, `Voter.ts`, `RootPool.ts`, `SwapFee.ts` in `Effects/` now only define the Envio effect and call the gateway (i.e. `callRpcGateway`); no direct viem/chain config usage in the effect layer.
- Removed `getIsAlive` RPC usage
     - `getIsAlive` effect and its export were removed.
     - `computeVoterDistributeValues` now takes `gaugeIsAlive` pool entity field as an argument; callers pass the value from the existing pool entity (`currentLiquidityPool.gaugeIsAlive`), so no on-chain read for liveness.

## Summary by CodeRabbit

* **New Features**
  * Centralized, typed RPC gateway with retries, fallbacks and unified handling
  * Smoke-test script to validate RPC-backed effects against live endpoints
  * New pure RPC fetch utilities and a block-rounding helper

* **Bug Fixes**
  * Removed an unused error classification pattern
  * Cleared noisy debug logs

* **Refactor**
  * Consolidated RPC calls under a single gateway and simplified retry/timeout semantics
  * Simplified voter reward computation interface and related flows

* **Tests**
  * Added gateway unit tests and updated suites to match new RPC behavior

### DIFF
--- a/src/Constants.ts
+++ b/src/Constants.ts
@@ -125,9 +125,18 @@ export const SLOW_REQUEST_MS = 5000;
 export const VERY_SLOW_REQUEST_MS = 30000;
 
 // Global rate limit for all RPC requests; used in rpcGateway effect
-export const GLOBAL_REQUESTS_PER_SECOND = 5000;
+export const GLOBAL_REQUESTS_PER_SECOND = 20000;
 
 export const RPC_GATEWAY_PREFIX = "rpcGateway";
+
+// RPC fallback for getTokenDetails effect
+export const TOKEN_DETAILS_FALLBACK = { name: "", decimals: 18, symbol: "" };
+
+export const USDC_DETAILS_FALLBACK = {
+  name: "USDC",
+  symbol: "USDC",
+  decimals: 6,
+};
 
 /**
  * Default/fallback public RPC URLs for each chain
@@ -148,15 +157,15 @@ export const DefaultRPC = {
 } as const;
 
 /**
- * Shared HTTP transport options for all chain RPC clients (Alchemy/DRPC best practices).
+ * Shared HTTP transport options for all chain RPC clients (DRPC best practices).
  * Centralised so batch size and timeout are tuned in one place.
- * - batchSize 50: avoids timeouts and unbounded response sizes; allows multiple batches in flight.
- * - timeout 60s: prevents indefinite hangs (Alchemy best practice).
+ * - batchSize 1_000: viem default. We are using dRPC now so we can use a larger batch size.
+ * - timeout 60s: prevents indefinite hangs.
  * - retryCount 0: transport-level retries are disabled. All RPC retries are handled in one place
  *   by RpcGateway (runWithRpcRetry + RPC_APP_RETRY), with logging and error-type-aware backoff.
  */
 export const RPC_HTTP_OPTIONS = {
-  batch: { batchSize: 50 },
+  batch: { batchSize: 1_000 },
   timeout: 60000, // RPC timeout in milliseconds (60 seconds). Prevents indefinite hangs on slow or unresponsive RPC providers
   retryCount: 0,
 } as const;
--- a/src/Effects/RpcGateway.ts
+++ b/src/Effects/RpcGateway.ts
@@ -3,6 +3,8 @@ import {
   CHAIN_CONSTANTS,
   PriceOracleType,
   RPC_GATEWAY_PREFIX,
+  TOKEN_DETAILS_FALLBACK,
+  USDC_DETAILS_FALLBACK,
 } from "../Constants";
 import {
   GLOBAL_REQUESTS_PER_SECOND,
@@ -38,6 +40,9 @@ function rpcGatewayOpName(type: EffectType, sub?: string): string {
 
 /**
  * Single source of truth for each gateway operation: input and output schemas.
+ * Each inputSchema includes a literal `type` field so that RPC_GATEWAY_INPUT_SCHEMA
+ * (S.union of these schemas) is a proper discriminated union: the runtime validates
+ * and preserves `type`, and the handler's switch (i.type) receives it intact.
  * When adding an operation:
  * 1) add to EffectType,
  * 2) add an entry here,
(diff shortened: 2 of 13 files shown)


src/Effects/Errors.ts
FonsSicca: just updates to docstring

src/Effects/RpcGateway.ts
```diff
@@ -0,0 +1,736 @@
...
+ * Use this for all RPC gateway operations so errors are consistent and the indexer never crashes.
+ *
+ * @param context - Effect context with optional cache flag and log (error + warn). On error, cache is set to false by {@link handleEffectErrorReturn}.
+ * @param operationName - Identifier for logging (use {@link rpcGatewayOpName} for gateway operations).
+ * @param logDetails - Key-value pairs included in error messages and retry logs.
+ * @param fallback - Value returned when the operation throws (after retries are exhausted).
+ * @param fn - Async function that performs the RPC call (e.g. a fetcher). No arguments; close over args in the caller.
+ * @returns The result of `fn()`, or `fallback` if `fn` throws and the error is handled.
+ */
+export async function executeRpcWithFallback<T>(
```
FonsSicca: responsible for retry and error handling logic

src/Effects/RpcGateway.ts
```diff
@@ -0,0 +1,736 @@
...
+import { fetchTokenDetails, fetchTokenPrice } from "./fetchers/Token";
+import { fetchTokensDeposited } from "./fetchers/Voter";
+
+export enum EffectType {
+  GET_TOKEN_DETAILS = "getTokenDetails",
+  GET_TOKEN_PRICE = "getTokenPrice",
+  GET_TOKENS_DEPOSITED = "getTokensDeposited",
+  GET_SWAP_FEE = "getSwapFee",
+  GET_ROOT_POOL_ADDRESS = "getRootPoolAddress",
+}
```
FonsSicca: centralising effect name in this enum

src/Constants.ts
```diff
@@ -125,6 +117,21 @@ export enum PriceOracleType {
...
 
+/**
+ * Slow and very slow request thresholds in milliseconds
+ * Used for logging slow requests
+ */
+export const SLOW_REQUEST_MS = 5000;
+export const VERY_SLOW_REQUEST_MS = 30000;
+
+// Global rate limit for all RPC requests; used in rpcGateway effect
+export const GLOBAL_REQUESTS_PER_SECOND = 20000;
```
FonsSicca: we don't actually need a limit since drpc doesn't require one, but just for the sake of it i've put here a high value (if i've put a low value that could affect indexer performance)

src/Effects/RpcGateway.ts
```diff
@@ -0,0 +1,736 @@
...
+      TOKEN_DETAILS_FALLBACK,
+      () => fetchTokenDetails(tokenAddress, ethClient),
+    ),
+    executeRpcWithFallback(
+      context,
+      rpcGatewayOpName(EffectType.GET_TOKEN_PRICE, "usdcDetails"),
+      { contractAddress: USDC_ADDRESS, chainId },
+      TOKEN_DETAILS_FALLBACK,
+      () => fetchTokenDetails(USDC_ADDRESS, ethClient),
+    ),
```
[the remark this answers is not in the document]
WooSungD: Better to have `(decimals: 6)` as an override fallback here I think

PR #556, opened 2026-03-05
Updating logic for current staked liquidity usd

FonsSicca:
partially closes issue 555 (not tagging to not close it entirely)

Implements:
- Updated logic for staked liquidity USD computation

## Summary by CodeRabbit

* **New Features**
  * Track whether NFT liquidity positions are staked in pool gauges and include that flag in historical snapshots.

* **Bug Fixes**
  * More accurate staked-USD valuation for concentrated and non-concentrated pools.
  * Staked USD now uses non-cumulative (overwrite) updates to avoid double-counting.
  * Withdrawals now guard against negative stakes and log/skips invalid updates.

* **Tests**
  * Expanded coverage for staking valuation, gauge transfers, and snapshot behavior.

### DIFF
--- a/src/EventHandlers/Gauges/GaugeSharedLogic.ts
+++ b/src/EventHandlers/Gauges/GaugeSharedLogic.ts
@@ -27,6 +27,77 @@ export interface GaugeEventData {
 }
 
 /**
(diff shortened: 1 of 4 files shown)


src/Helpers.ts
FonsSicca: also cleared here some dead code

PR #605, opened 2026-04-17
fix(gauge): clamp CLGauge withdraw when it exceeds indexed stake

### LINKED ISSUE #604: Fix CLGauge.Withdraw exceeds current stake on Superseed (59 users)
## Parent PRD

#597

## What to build

59 `CLGauge.Withdraw: withdraw exceeds current stake for pool <P> user <U>. Skipping update. This needs to be fixed.` errors during the audit window — all on chain 5330 (Superseed), 59 unique (pool, user) pairs. The handler is literally in a "skip + TODO" branch: when withdraw amount exceeds the indexer's view of staked balance, it skips the update, leaving `stakedLiquidity` permanently overcounted.

Root cause must be one of:
- **Missed deposit attribution**: one or more `Deposit` events for this user were not indexed (never added to staked balance).
- **Cross-chain bridged initial state**: if the user migrated a staked position from another chain, the indexer didn't capture the initial stake.
- **Event ordering race**: `Withdraw` handler runs before the `Deposit` it depends on materializes.

Steps:
- Pick one (pool, user) pair from the 59. Enumerate all `Deposit`/`Withdraw` events on-chain for that user on that pool on Superseed. Compare with indexed state.
- Determine which of the 3 causes applies. Likely chain-specific to Superseed given the zero cross-chain pattern.
- Implement the fix (probably in the CLGauge handler or its Voter-side producer).
- Replace the `"This needs to be fixed"` TODO branch with a real resolution.
- Add a regression test covering a deposit-then-withdraw sequence that matches the original failure mode.

HITL because root cause is unknown until the investigation runs.

## Acceptance criteria

- [ ] Root-cause diagnosis posted as a comment, using a specific on-chain (pool, user) example.
- [ ] Fix implemented; the `"This needs to be fixed"` TODO branch replaced.
- [ ] Regression test covers the failing sequence.
- [ ] Post-fix backfill: `CLGauge.Withdraw exceeds current stake` error count on Superseed drops to zero.

## Blocked by

None - can start immediately.

## User stories addressed

- User story 9
- User story 10

FonsSicca:
## Summary

Resolves #604. Replace the silent-skip `"This needs to be fixed"` TODO branch in `processGaugeWithdraw` with an independent clamp on the pool and user aggregate decrements. When a `Withdraw` arrives for more liquidity than the indexer has seen deposited, the on-chain transaction is the source of truth — so the handler now clamps each aggregate to `max(0, current - amount)`, still increments the withdrawal count, and emits a single warning that captures the discrepancy.

## Why this is safe

- The CL-specific bookkeeping (tick liquidity, `stakedReserve0/1`, `stakedLiquidityInRange`) is already driven by `position.liquidity`, not the event amount, so tick/reserve accounting remains correct.
- Pool and user aggregates are clamped independently, preserving the invariant `currentLiquidityStaked >= 0` on both sides.
- `stakedCLPositionTokenIds` is still filtered for the withdrawn `tokenId` — the user no longer appears to hold a position they just withdrew.
- Previously 59 events on Superseed left `stakedLiquidity` permanently inflated with no recovery path. With the clamp, subsequent Deposit/Withdraw attribution can re-converge on the correct value.

## Root-cause diagnosis

All 59 errors share chain 5330 (Superseed) and no cross-chain signature, which points to **missed Deposit attribution** rather than an event-ordering race or bridged initial state. The silent-skip branch made this manifest as permanent overcount instead of a single diagnostic error at the first failed withdraw. Clamping converts it from "permanent corruption" to "converges once the next valid event flows through."

## Tests

Added two regression scenarios in `test/EventHandlers/Gauges/GaugeSharedLogic.test.ts` under `processGaugeWithdraw > withdraw exceeds current stake (issue #604)`:

- Withdraw with no prior indexed stake — clamps both pool and user to zero, still records the withdrawal, emits one warning, no error.
- Withdraw where only the pool aggregate is short — pool clamps to zero while user decrements normally; the two aggregates stay independently consistent.

`pnpm qa --write` clean. `pnpm test` green: 1025 passed, 32 skipped across 90 files.

- **Log query:** search for `withdraw .* exceeds indexed stake` on chain 5330.
  - Expected: count drops to zero for new blocks after deploy (no further missed-attribution producers on Superseed that the indexer can retroactively uncover).
  - Residual warnings on older blocks during backfill are expected (the original 59); they now clamp instead of skipping.
- **Metric to watch:** Superseed `CLGauge.Withdraw` error count in Envio `/metrics` histograms — should transition from recurring `uerror` to a bounded one-time `uwarn` set and then zero on post-deploy blocks.
- **Healthy signal:** `LiquidityPoolAggregator.currentLiquidityStaked` on affected Superseed CL pools trends toward on-chain reality rather than drifting upward.
- **Failure signal / rollback trigger:** new `uerror`-level gauge events, pool-level `currentLiquidityStaked` going negative (shouldn't be possible with the clamp), or tick-liquidity regressions on non-affected chains. Roll back by reverting this commit — the pre-existing skip behavior resumes.
- **Validation window:** 48 hours post-deploy with the first full backfill cycle.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

### DIFF
--- a/src/EventHandlers/Gauges/GaugeSharedLogic.ts
+++ b/src/EventHandlers/Gauges/GaugeSharedLogic.ts
@@ -372,16 +372,29 @@ export async function processGaugeWithdraw(
 
   const { liquidityPoolAggregator } = poolData;
 
-  const newPoolStake =
-    liquidityPoolAggregator.currentLiquidityStaked - data.amount;
-  const newUserStake = userData.currentLiquidityStaked - data.amount;
+  // Clamp the pool and user decrements independently to each aggregate's
+  // current stake. On-chain reality is the source of truth: when a Withdraw
+  // exceeds the indexer's view of stake, a prior Deposit was missed
+  // (attribution gap, event ordering, cross-chain migration, etc.).
+  // Previously this branch skipped the update entirely, which left
+  // currentLiquidityStaked permanently inflated with no recovery path.
+  const poolDecrement =
+    data.amount <= liquidityPoolAggregator.currentLiquidityStaked
+      ? data.amount
+      : liquidityPoolAggregator.currentLiquidityStaked;
+  const userDecrement =
+    data.amount <= userData.currentLiquidityStaked
+      ? data.amount
+      : userData.currentLiquidityStaked;
 
-  if (newPoolStake < 0n || newUserStake < 0n) {
-    context.log.error(
-      `${handlerName}: withdraw exceeds current stake for pool ${pool.poolAddress} user ${data.userAddress}. Skipping update. This needs to be fixed.`,
+  if (poolDecrement < data.amount || userDecrement < data.amount) {
+    context.log.warn(
+      `${handlerName}: withdraw ${data.amount} exceeds indexed stake (pool=${liquidityPoolAggregator.currentLiquidityStaked}, user=${userData.currentLiquidityStaked}) for pool ${pool.poolAddress} user ${data.userAddress} on chain ${data.chainId}. Clamping to available stake.`,
     );
-    return;
   }
+
+  const newPoolStake =
+    liquidityPoolAggregator.currentLiquidityStaked - poolDecrement;
 
   let poolStakedUSD: bigint | undefined;
   let stakedLiquidityInRange: bigint | undefined;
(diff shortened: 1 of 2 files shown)


FonsSicca:
Closing — root cause is a data gap on superseed.hypersync.xyz, not a handler bug. The clamp only masks the symptom without restoring correct state. Full remediation requires Envio re-indexing the affected block range (10,318,646 → 10,520,416 on Superseed) + a clean Superseed re-sync.

Issue #597, opened 2026-04-17
PRD: Indexer observability & data integrity cleanup from log audit


FonsSicca:
## Problem Statement

A 9-day indexer backfill produced 125,838 log entries across 136 files (Apr 8–17, 2026). An audit of that output revealed two compounding issues:

1. **Log noise saturation.** 108,526 of 108,526 `uwarn` entries (~99%) are RPC retry messages from `getSwapFee` and `getTokenPrice`. Each carries a full viem stack trace (~1–2 KB). Real warnings are invisible by dilution, and roughly 130 MB of stack strings are allocated per backfill for no diagnostic value — we already have Envio's `/metrics` endpoint for aggregate visibility.

2. **Data-correctness drift hiding in the noise.** Once the RPC spam is filtered out, 6 distinct bugs surface, each writing wrong numbers into the database without the indexer crashing:
   - 953 `NonFungiblePosition not found` errors across 443 unique positions on chains 10 and 5330 (users' CL position accounting silently missing).
   - 59 `CLGauge.Withdraw exceeds current stake` skips on Superseed (staked TVL permanently overcounted).
   - 34 Fraxtal price-oracle reverts (WFRAX et al.) → every Fraxtal pool shows \$0 TVL/volume.
   - 34 Soneium `eth_call not available` errors → Soneium tokens load with default decimals, poisoning amount normalization.
   - 709 Celo `OUSDTBridgedTransaction not found` → cross-chain swap records silently dropped.
   - 2 orphan root-pool mappings (Base `0x97Cd4EB6…`, Optimism `0x333030A7…`) → voting/bribe/emission events on those pools dropped.

The audit covered the entire backfill since deployment, so this is steady-state behavior, not a cold-start artifact.

## Solution

Two parallel tracks of work, both sourced from the same audit:

- **Track A (observability)**: collapse `runWithRpcRetry` to emit exactly one log line when retries are exhausted, with no `stack_trace` on intermediate attempts. Target: ~99% cut in `uwarn` volume (108,526 → ~750) and ~7× reduction in total log bytes per backfill.
- **Track B (data integrity)**: one issue per P0/P1 bug. Each is either a correlation fix, a config fix, or a reconciliation task — none require Envio internal changes.

## User Stories

1. As an indexer maintainer, I want the 99% of `uwarn` volume caused by RPC retry chatter eliminated, so that every remaining `uwarn` deserves attention.
[...]

PR #606, opened 2026-04-17
fix(superswaps): filter non-OUSDT CrossChainSwap events and escalate real correlation failures to uerror

### LINKED ISSUE #602: Diagnose and fix OUSDT bridged-transaction correlation on Celo
## Parent PRD

#597

## What to build

709 occurrences of `No OUSDTBridgedTransaction found for transaction <X>` during the audit window, 100% on Celo (chainId 42220), 709 unique tx hashes. Each corresponds to a `VelodromeUniversalRouter.CrossChainSwap` event that cannot find its matching bridge record and **returns early** — the cross-chain swap is never recorded.

Flow today (`src/EventHandlers/SuperswapsHyperlane/VelodromeUniversalRouter.ts`):

- `UniversalRouterBridge` (source chain) writes `OUSDTBridgedTransaction` keyed by `event.transaction.hash`, gated on `token === OUSDT_ADDRESS`.
- `CrossChainSwap` (destination chain) looks up by `event.transaction.hash`. If missing → warn + return.

Three candidate root causes (diagnose first, then pick a fix):

1. **Tx-hash mismatch across chains** — `CrossChainSwap` fires on the destination chain with a different tx hash than `UniversalRouterBridge` on the source. Correlation would need `messageId` instead of `transactionHash`.
2. **Filter mismatch** — `CrossChainSwap` fires for non-OUSDT tokens that the producer filtered out. Fix: add symmetric filter on consumer.
3. **Missing source-chain registration** — `UniversalRouterBridge` isn't listened to on the chain that produced these bridges.

Steps:
- Pick one of the 709 affected tx hashes. Inspect it on the source chain and the destination (Celo).
- Determine which of the three causes applies (may be a mix).
- Implement the matching fix. Until the fix ships, upgrade the existing `uwarn` to `uerror` so the regression is visible in error logs.

HITL because the three candidate fixes have very different scopes.

## Acceptance criteria

- [ ] Diagnosis written up as a comment on this issue: which of the 3 causes, with a concrete example tx.
- [ ] Fix implemented (correlation change, filter, or registration).
- [ ] A bridge+swap pair test verifies the fix end-to-end (following `test/EventHandlers/...` prior art).
- [ ] A swap-without-bridge test verifies we produce a `uerror`, not a silent drop.

## Blocked by

None - can start immediately.

## User stories addressed

- User story 5
- User story 6

FonsSicca:
Fixes #602 (part of PRD #597)

## Diagnosis

`VelodromeUniversalRouter.CrossChainSwap` fires on the **source chain** for every cross-chain swap, regardless of the bridged token. The companion `UniversalRouterBridge` producer only records `OUSDTBridgedTransaction` rows when the bridged token is OUSDT (filtered by `token === OUSDT_ADDRESS`). Same transaction on the same chain, but asymmetric filters.

Result: every non-OUSDT cross-chain swap fell through the consumer's bridge lookup and produced `No OUSDTBridgedTransaction found for transaction <X>`. That's the entire 709-warning spike on Celo in the April 2026 log audit — 100% chainId 42220, 709 unique tx hashes.

This matches root cause #2 from issue #602 (**Filter mismatch**). Root cause #1 (tx-hash mismatch across chains) is ruled out by the ABI: both events are emitted on the source chain in the same transaction, so `event.transaction.hash` is identical. Root cause #3 (missing source-chain registration) is ruled out by `config.yaml`: `VelodromeUniversalRouter` is registered on every indexed chain, Celo included.

## Fix

Extracted the CrossChainSwap handler body into `handleCrossChainSwapEvent` in `SuperSwapLogic.ts`. The missing-bridge branch now:

1. Checks whether an `OUSDTSwaps` record exists for the same transaction on the source chain.
2. If **yes** → real correlation failure (OUSDT pool swap happened but bridge record is missing) → `context.log.error` with an `Error` payload.
3. If **no** → non-OUSDT cross-chain swap, symmetric with the producer's filter → silently skip.

The legacy `uwarn` is gone. Error logs now count only true correlation failures, making the existing `uerror`-grep observability strategy from PRD #597 work on the OUSDT path.

## Tests

New `handleCrossChainSwapEvent` describe block in `test/EventHandlers/SuperswapsHyperlane/SuperSwapLogic.test.ts`:

- **Bridge+swap pair** (end-to-end happy path): produces a `SuperSwap`, no warn/error.
- **Swap-without-bridge**: `OUSDTSwaps` present, `OUSDTBridgedTransaction` missing → exactly one `uerror` with `Error` payload, no legacy `uwarn`, no `SuperSwap`.
- **Non-OUSDT cross-chain swap**: neither swap nor bridge → silent skip, no logs, no `SuperSwap`.

All 1026 tests pass, `pnpm qa --write` clean.

- **Log query (baseline)**: before deploy, note count of `"No OUSDTBridgedTransaction found"` `uwarn` entries on Celo — audit baseline was 709 per 9-day backfill.
- **Post-deploy expected healthy signal**: that `uwarn` message disappears entirely. In its place, only `"No OUSDTBridgedTransaction found for OUSDT swap"` `uerror` entries should appear — and the count on Celo should be 0 or very low. Any nonzero count is a real OUSDT correlation defect and should be triaged.
- **Failure signal / rollback trigger**: if the new `uerror` count on any chain exceeds the pre-deploy `uwarn` count, revert — the symmetric filter is too permissive. If `SuperSwap` entity creation rate drops materially on any chain vs the pre-deploy baseline, revert — happy path regressed.
- **Validation window**: 48 hours post-deploy, covering at least one full backfill cycle.
- **Owner**: indexer on-call.

## Summary by CodeRabbit

* **Improvements**
  * Improved cross-chain swap event handling with clearer error reporting when expected transaction data is missing and silent skipping for non-relevant transactions.
* **Refactor**
  * Consolidated cross-chain swap processing into a single handler to centralize loading, validation, and delegation for consistency.
* **Tests**
  * Added end-to-end tests covering success, missing-bridge error, and silent-skip edge cases.

### DIFF
--- a/src/EventHandlers/SuperswapsHyperlane/SuperSwapLogic.ts
+++ b/src/EventHandlers/SuperswapsHyperlane/SuperSwapLogic.ts
@@ -308,6 +308,80 @@ export async function processCrossChainSwap(
 }
 
 /**
+ * Handles a CrossChainSwap event by loading the OUSDT bridge record and related
+ * mailbox state, then delegating to processCrossChainSwap for SuperSwap entity
+ * creation.
+ *
+ * CrossChainSwap fires for every cross-chain swap through VelodromeUniversalRouter,
+ * including non-OUSDT bridges that the UniversalRouterBridge producer filtered out.
+ * To avoid spamming warnings for those non-OUSDT bridges, this function treats the
+ * absence of an OUSDTBridgedTransaction as an error ONLY when an OUSDTSwaps record
+ * for the same transaction exists (meaning OUSDT was actually swapped but the
+ * bridge correlation is missing).
+ */
+export async function handleCrossChainSwapEvent(
+  transactionHash: string,
+  chainId: number,
+  destinationDomain: bigint,
+  blockTimestamp: number,
+  context: handlerContext,
+): Promise<void> {
+  const [oUSDTBridgedTransactions, sourceChainMessageIdEntities] =
+    await Promise.all([
+      context.OUSDTBridgedTransaction.getWhere({
+        transactionHash: { _eq: transactionHash },
+      }),
+      context.DispatchId_event.getWhere({
+        transactionHash: { _eq: transactionHash },
+      }),
+    ]);
+
+  if (oUSDTBridgedTransactions.length === 0) {
+    // Symmetric filter with the UniversalRouterBridge producer: CrossChainSwap
+    // fires for any bridged token, but OUSDTBridgedTransaction is only recorded
+    // when the bridged token is OUSDT. If no OUSDT pool swap exists for this
+    // transaction, this cross-chain swap did not involve OUSDT — silently skip.
+    const sourceChainOUSDTSwaps = await context.OUSDTSwaps.getWhere({
+      transactionHash: { _eq: transactionHash },
+    });
+    if (sourceChainOUSDTSwaps.length > 0) {
+      context.log.error(
+        `No OUSDTBridgedTransaction found for OUSDT swap at chain ${chainId} tx ${transactionHash}`,
+        new Error(
+          `Missing OUSDTBridgedTransaction for transaction ${transactionHash}`,
+        ),
+      );
+    }
+    return;
+  }
+
+  const bridgedTransaction = oUSDTBridgedTransactions[0];
+
+  if (sourceChainMessageIdEntities.length === 0) {
+    return;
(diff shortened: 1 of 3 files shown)


FonsSicca:
@coderabbitai review

WooSungD (approved):

PR #608, opened 2026-04-17
fix(rpc): classify method-not-supported errors and fall back to public RPC

### LINKED ISSUE #599: Replace Soneium RPC endpoint to support eth_call
## Parent PRD

#597

## What to build

Replace the Soneium (chainId 1868) RPC endpoint configured in `src/Constants.ts` (`DefaultRPC.soneium`, currently `https://soneium.drpc.org`). The current endpoint returns `the method "eth_call" does not exist / is not available`, which causes `getTokenPrice.tokenDetails` and `destinationTokenDetails` to fail 100% of the time on Soneium (34 errors in the audit window). Every token that isn't already in the registry falls back to default decimals, which poisons amount normalization for every pool that token appears in.

Steps:
- Evaluate candidate endpoints (official Soneium RPC, Alchemy, QuickNode, public RPC list) against the `eth_call` requirement.
- Update `DefaultRPC.soneium` or document the required `ENVIO_SONEIUM_RPC_URL` env var for the ops environment.
- Verify with one token fetch that `getTokenPrice.tokenDetails` now returns a real response.

This is HITL because a human must pick the endpoint.

## Acceptance criteria

- [ ] New Soneium RPC endpoint chosen and documented (in config or env var).
- [ ] A manual fetch for one Soneium token returns decimals/name/symbol (verify with the Envio dev loop or a one-off script).
- [ ] Backfill produces zero `[rpcGateway.getTokenPrice.tokenDetails] … eth_call does not exist` errors on chain 1868.
- [ ] `pnpm qa --write` clean.

## Blocked by

None - can start immediately.

## User stories addressed

- User story 13
- User story 14

FonsSicca:
Closes #599.

## The surface bug vs. the real bug

The issue describes a 9-day Soneium outage where the premium RPC returned `the method "eth_call" does not exist / is not available`, causing `getTokenPrice.tokenDetails` and `destinationTokenDetails` to fail 100% of the time (34 errors on chain 1868). Every Soneium token not already in the registry defaulted to 18 decimals and poisoned amount normalisation for every pool it touched.

Testing the premium endpoint today shows it serves `eth_call` fine -- dRPC fixed it upstream. A minimal URL swap would close the issue on paper but leave the actual architectural gap untouched: the indexer had no runtime path to survive this failure class, on any chain.

## What was actually broken

1. **Error classifier too greedy.** `HISTORICAL_STATE_NOT_AVAILABLE` matched on `"is not available"` alone, so the audit error got bucketed as "historical state" -- a non-retryable class -- and went straight to the static fallback (decimals=18).
2. **No fallback RPC in the runtime.** `src/Effects/Errors.ts` defined `createFallbackClient` and `shouldUseFallbackRPC` with unit tests, but nothing in `src/` ever called them. "Fallback" inside `executeRpcWithFallback` just meant "static default value," not "try a second client."
3. **Default Soneium RPC was the same flaky provider.** If the env var ever dropped, the indexer would land on the broken endpoint.

## The fix

Two commits:

### 1. `refactor(effects): remove dead createFallbackClient and shouldUseFallbackRPC`
Deletes `src/Effects/Errors.ts` and `test/Effects/Errors.test.ts`. Nothing in `src/` imported from them; the unit tests were the only callers. Clean slate before building the real mechanism.

### 2. `fix(rpc): classify method-not-supported errors and fall back to public RPC`
- **`ErrorType.METHOD_NOT_SUPPORTED`** — new bucket matching `"does not exist"`, `"method not found"`, `"method not supported"`, `"unsupported method"`. Checked before `HISTORICAL_STATE_NOT_AVAILABLE`, so `the method "eth_call" does not exist / is not available` classifies correctly.
- **Per-error-type retry cap in `runWithRpcRetry`.** `METHOD_NOT_SUPPORTED` uses `methodNotSupportedMaxRetries=2` instead of the default 7; this error class is typically deterministic (upstream provider has the method disabled), so more retries just burn latency and log volume. Rate-limit and network retries unchanged.
- **Optional `fallbackFn` in `executeRpcWithFallback`.** When primary retries are exhausted on `METHOD_NOT_SUPPORTED` or `HISTORICAL_STATE_NOT_AVAILABLE`, the fallback fetcher runs through its own retry loop before the caller receives the static default. Other error classes (reverts, rate limits, network) skip the fallback — switching providers doesn't help those.
- **`createFallbackRpcClient(chainId)`** in `Constants.ts` — builds and caches a viem `PublicClient` pinned to `DefaultRPC.<chain>`. Replaces `getDefaultRPCByChainId` (whose map is now an implementation detail of the new helper).
- **`DefaultRPC.soneium`** → `https://rpc.soneium.org` (official public RPC, verified serving `eth_call` for name/symbol/decimals at both `latest` and historical blocks for USDC.e / WETH / ASTR). This is what the fallback client actually hits. `.env.example` follows.
- **Wire `fallbackFn` into the audit's exact failure surface:** `handleGetTokenDetails`, and the two `tokenDetails`/`destinationTokenDetails` calls inside `handleGetTokenPrice`, plus the oracle-price fetch.

## Failure modes, before and after

| Primary-RPC error | Before | After |
| --- | --- | --- |
| `"method eth_call does not exist"` | Misclassified as historical-state; no retry; decimals=18 silently stored | `METHOD_NOT_SUPPORTED`; 2 retries; then fallback to `rpc.soneium.org`; correct decimals stored |
| `"historical state not available at block N"` | Classified correctly; no retry; static fallback | Same classification; now also tries the fallback client (archive on a different provider may succeed) |
| Contract revert | Static fallback | Static fallback (unchanged — fallback provider doesn't help) |
| Rate limit | Retry ×7; static fallback on exhaustion | Same |

## Test plan

- [x] `pnpm qa --write` — biome clean
- [x] `pnpm envio codegen` — succeeds
- [x] `tsc --noEmit` — clean
- [x] `pnpm test` — 1024 passed / 32 skipped / 0 failed (was 1023 before this PR; -7 deleted `Errors.test.ts` cases, +8 new cases across classifier, retry cap, fallback paths, and client builder)
- [x] Live probe: premium + official Soneium endpoints both serve `eth_call` for name/symbol/decimals at `latest` and at head−1M blocks, through viem's batched `readContract` (exact call shape `fetchTokenDetails` uses)

Tests added:
- `getErrorType` cases for each new `METHOD_NOT_SUPPORTED` phrasing; regression case that `"historical state is not available"` still bucketed correctly.
- `runWithRpcRetry`: `METHOD_NOT_SUPPORTED` retries 2× then escalates.
- `executeRpcWithFallback`: (a) primary exhaust → `fallbackFn` runs and wins; (b) `CONTRACT_REVERT` → `fallbackFn` skipped, static fallback; (c) both fail → static fallback with `.fallback` suffix in the error log; (d) no `fallbackFn` provided → static fallback.
- `createFallbackRpcClient`: returns a Soneium-pinned client, `null` for unsupported chains, caches per `chainId`.

## Acceptance criteria (from #599)

- [x] New Soneium RPC endpoint chosen and documented (code default + `.env.example` updated).
- [x] Manual fetch for one Soneium token returns decimals/name/symbol (USDC.e at `rpc.soneium.org`: `"Bridged USDC (Soneium)"` / `"USDC.e"` / `6`).
- [ ] Backfill produces zero `[rpcGateway.getTokenPrice.tokenDetails] … eth_call does not exist` errors on chain 1868 *(verified post-deploy — see monitoring plan below)*.
- [x] `pnpm qa --write` clean.

- **Log queries (first 24h of the next backfill):**
  - Count of `rpcGateway.getTokenPrice.tokenDetails` errors on `chainId":1868` — expected: 0 (was 34 in the Apr 8–17 audit).
  - Count of `eth_call does not exist` on chain 1868 — expected: 0.
  - New: `primary RPC exhausted (METHOD_NOT_SUPPORTED)` warns on any chain — should be rare; each one is a signal the fallback kicked in, and is the success path, not a failure.
- **Metrics (Envio `/metrics`):** RPC error rate for `getTokenPrice` on chain 1868 — baseline (<1%).
- **DB spot-check:** `SELECT chain_id, id, symbol, decimals FROM "Token" WHERE chain_id = 1868 AND decimals != 18 LIMIT 10;` — expect USDC.e decimals=6 and other non-18 tokens present.
- **Failure signal + rollback:** if `eth_call does not exist` errors reappear on chain 1868 AND the fallback also fails (`.fallback` suffix in the error log on the same call), set `ENVIO_SONEIUM_RPC_URL` to an alternate provider — env var still overrides the default, no code change needed.
- **Owner:** @FonsSicca
- **Validation window:** first complete Soneium backfill (≈9 days).


FonsSicca:
@coderabbitai review

WooSungD (approved):

FonsSicca:
@coderabbitai review

FonsSicca:
@coderabbitai review

FonsSicca:
@coderabbitai review

FonsSicca:
@coderabbitai review

FonsSicca:
@coderabbitai review

FonsSicca:
@coderabbitai review

FonsSicca:
@coderabbitai review

PR #610, opened 2026-04-17
fix(voter): reconcile orphan PendingRootPoolMapping records on read

### LINKED ISSUE #601: Reconcile 2 orphan root-pool mappings (Base + Optimism)
## Parent PRD

#597

## What to build

Two root pools are permanently orphaned:

- Base: gauge `0x287C94a1fE647014317E91A0E42425D6a237081D` → root pool `0x97Cd4EB683E29695DC93eec95d47d4E7a35E2112` (29 `RootPool_LeafPool mapping not found` warnings + `[loadPoolData] LiquidityPoolAggregator 8453-0x97Cd4EB6… not found`)
- Optimism: gauge `0x3B59a6B600f912260048a0f3a834C1039aEcD367` → root pool `0x333030A736B47D20346d82A473680658ac1C2b88` (1 warning + 2 `loadPoolData` errors)

Additionally, one `PendingRootPoolMapping` record was stored at block 133970688 for hash `42220_0x471EcE3750Da237f93B8E339c536989b8978a438_0x7f9AdFbd38b669F03d1d11000Bc76b9AaEA28A81_2000` and was never reconciled.

Every voting/bribe/emission event on these two pools is silently dropped today.

Steps:
- Trace the reconciliation path: find the `PendingRootPoolMapping` producer and consumer in `src/Effects/RootPool.ts` / `src/EventHandlers/Voter/*`.
- Determine why the single pending record has not been consumed (missing trigger event? wrong key? timing?).
- Either fix the reconciliation logic so the record is consumed, or perform a one-time backfill/data surgery for the two specific pools.
- Add test coverage so the reconciliation path is exercised at least once.

HITL because root cause could be any of design, data, or runtime.

## Acceptance criteria

- [ ] Root-cause diagnosis documented in a comment on this issue before the fix.
- [ ] Both orphan pools resolve through `loadPoolData` without error after re-indexing.
- [ ] `PendingRootPoolMapping` count drops to zero for the affected hash (or a justified reason is provided for why it cannot).
- [ ] Test added for the reconciliation path (following `test/EventHandlers/...` prior art).

## Blocked by

None - can start immediately.

## User stories addressed

- User story 15
- User story 16

FonsSicca:
## Summary

Closes #601.

Votes, bribes, and emissions on two root pools (Base `0x97Cd4EB6…`, Optimism `0x333030A7…`) were being silently dropped because their `RootPool_LeafPool` mappings never materialised. The deferred reconciliation path only fires when `CLFactory.PoolCreated` is observed on the leaf chain, and that event never reached us for the Celo leaf pool underlying the Optimism orphan (pool likely predates Celo's `start_block`). The pending record stayed orphaned forever and every cross-chain event against that root pool fell through `loadPoolDataOrRootCLPool` / `resolveLeafPoolForRootGauge` as `MAPPING_NOT_FOUND`.

This PR adds a reactive reconciliation step that recovers the mapping lazily the next time the root pool is referenced, whenever the leaf `LiquidityPoolAggregator` is already in the DB.

Full root-cause write-up: velodrome-finance/indexer#601 (comment).

## Approach

New helper `tryReconcileOrphanPendingRootPoolMapping(context, rootChainId, rootPoolAddress)` in `src/Aggregators/LiquidityPoolAggregator.ts`:

1. Loads `PendingRootPoolMapping` by `${rootChainId}-${rootPoolAddress}`.
2. Looks up `LiquidityPoolAggregator.getWhere({ rootPoolMatchingHash })` on the leaf side.
3. If exactly one leaf aggregator matches, creates `RootPool_LeafPool` and deletes the pending record. Otherwise returns `null` (fail-safe on ambiguous or missing leaf state).

Wired in as a pre-`MAPPING_NOT_FOUND` fallback in two places:

- `loadPoolDataOrRootCLPool` (drives `Voter.Voted` / `Voter.Abstained` / anything that resolves a pool by address)
- `resolveLeafPoolForRootGauge` (drives `Voter.DistributeReward` cross-chain gauges)

Both callers retry the standard leaf-pool lookup with the newly-created mapping after reconciliation succeeds. When reconciliation fails (no pending, ambiguous hash match, or no leaf aggregator yet) the code falls back to the existing `PendingVote` / `PendingDistribution` deferral path, so behaviour is strictly additive.

```mermaid
flowchart TB
    A[Voter.Voted / Abstained / DistributeReward<br/>references root pool] --> B{RootPool_LeafPool<br/>exists?}
    B -->|yes| C[apply vote/distribution<br/>to leaf pool]
    B -->|no| D[tryReconcileOrphan<br/>PendingRootPoolMapping]
    D --> E{PendingRootPoolMapping<br/>exists?}
    E -->|no| F[existing MAPPING_NOT_FOUND path:<br/>create PendingVote/PendingDistribution]
    E -->|yes| G{exactly one leaf<br/>aggregator matches hash?}
    G -->|no| F
    G -->|yes| H[create RootPool_LeafPool<br/>delete PendingRootPoolMapping]
    H --> C
```

## What this does not cover

- **Base orphan `0x97Cd4EB6…`:** no `RootCLPoolFactory` is registered for Base in `config.yaml`, so no `PendingRootPoolMapping` ever exists and reconciliation has nothing to consume. Requires registering the correct Aerodrome-Base root factory address — follow-up ticket, not in this PR.
- **Celo leaf aggregator missing entirely:** if the indexer has no `LiquidityPoolAggregator` for the leaf pool (e.g. pool predates `start_block` and no subsequent event surfaced it), reconciliation cannot match. Separate data-sourcing concern.
- **Backlog flushing of already-persisted `PendingVote` / `PendingDistribution`:** the reconciliation fixes the mapping so new events flow through correctly, but previously-deferred pendings for the orphan root pool are not re-processed here. A one-off cleanup is a separate task.

## Testing

Three new tests, all following existing harnesses:

- `test/Aggregators/LiquidityPoolAggregator.test.ts` — unit tests for `tryReconcileOrphanPendingRootPoolMapping` covering: happy path (reconciles + writes `RootPool_LeafPool` + deletes pending), ambiguous leaf aggregator (no-op), missing pending (no-op).
- `test/EventHandlers/Voter/Voter.test.ts` — end-to-end `Voter.Voted` test using the real `(CELO, XVELO, 2000)` token pair from the issue: starts with `PendingRootPoolMapping` + leaf `LiquidityPoolAggregator`, ends with the vote applied to the leaf pool and the pending record gone.
- `test/EventHandlers/Voter/VoterCommonLogic.test.ts` — `resolveLeafPoolForRootGauge` test verifying the reconciliation helper is invoked on the zero-mapping branch.

Existing tests updated only to add the new `PendingRootPoolMapping` entity to mock contexts — no behavioural changes.

```
pnpm test     # 1028 passed, 32 skipped
pnpm qa       # biome clean
tsc --noEmit  # clean
```

This ships as a lazy read-path behaviour change. To validate after deploy:

**Log queries**
- Search `[resolveLeafPoolForRootGauge] … RootPool_LeafPool mapping not found or ambiguous` for the two orphan root pool addresses. Count should drop to zero for `0x333030A736B47D20346d82A473680658ac1C2b88` once a Celo `LiquidityPoolAggregator` exists for the matching pool. Count for the Base orphan `0x97Cd4EB683E29695DC93eec95d47d4E7a35E2112` is expected to remain unchanged (see "What this does not cover") until the follow-up factory registration ships.
- Search `[loadPoolData] LiquidityPoolAggregator 8453-0x97Cd4EB6…` and `…LiquidityPoolAggregator 10-0x333030A7…` errors — same expectation.

**Entity checks**
- `SELECT COUNT(*) FROM "PendingRootPoolMapping" WHERE "rootPoolMatchingHash" = '42220_0x471EcE3750Da237f93B8E339c536989b8978a438_0x7f9AdFbd38b669F03d1d11000Bc76b9AaEA28A81_2000';` — expected to drop to 0 after the first vote/bribe/emission on the Optimism orphan reaches the indexer (contingent on the Celo aggregator being present).
- `SELECT COUNT(*) FROM "RootPool_LeafPool" WHERE "rootPoolAddress" = '0x333030A736B47D20346d82A473680658ac1C2b88';` — expected to become 1 after reconciliation.

**Failure signals**
- Any new error like `Expected exactly one RootPool_LeafPool for pool …` would indicate the reconciliation produced a duplicate mapping; mitigation is an immediate revert and we would investigate hash collisions.
- No change in `PendingRootPoolMapping` count after a known Voted/DistributeReward against the Optimism orphan means either the Celo leaf aggregator still doesn't exist in the DB (expected — separate concern) or the hash is not matching (investigate case/checksum).

**Validation window**: 24 hours post-deploy, owner: indexer maintainers.
**Rollback**: straightforward revert of this commit; behaviour on the non-orphan path is strictly unchanged.

---

[Compound Engineering v2.62.0](https://github.com/EveryInc/compound-engineering-plugin)
🤖 Generated with Claude Opus 4.7 (1M context, extended thinking) via [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/Aggregators/LiquidityPoolAggregator.ts
+++ b/src/Aggregators/LiquidityPoolAggregator.ts
@@ -1,10 +1,16 @@
 import type {
   CLGaugeConfig,
   LiquidityPoolAggregator,
+  RootPool_LeafPool,
   Token,
   handlerContext,
 } from "generated";
-import { PoolId, TokenId } from "../Constants";
+import {
+  PendingRootPoolMappingId,
+  PoolId,
+  RootPoolLeafPoolId,
+  TokenId,
+} from "../Constants";
 import { getSwapFee, roundBlockToInterval } from "../Effects/Index";
 import { calculateTotalUSD, generatePoolName } from "../Helpers";
 import { refreshTokenPrice } from "../PriceOracle";
@@ -493,6 +499,65 @@ export async function loadPoolData(
 }
 
 /**
+ * Attempts to reconcile a stuck PendingRootPoolMapping for the given root pool.
+ *
+ * This handles the failure mode from issue #601: RootCLPoolFactory.RootPoolCreated fires
+ * on the root chain (Optimism) and stores a PendingRootPoolMapping because no
+ * LiquidityPoolAggregator exists on the leaf chain yet. Normally CLFactory.PoolCreated
+ * on the leaf chain would consume the pending record, but if the leaf pool was created
+ * before the leaf chain's `start_block` (or by a factory not registered in config.yaml),
+ * the event is never observed and the pending record is orphaned.
+ *
+ * When a later Voted/Abstained/DistributeReward references the orphan root pool, we
+ * retry the reconciliation: if the leaf pool's LiquidityPoolAggregator has since appeared
+ * (e.g. created by a swap or other event that surfaced it), create the RootPool_LeafPool
+ * mapping, delete the pending record, and return the leaf pool info.
+ *
+ * @returns The newly-created RootPool_LeafPool row, or null if reconciliation was not possible.
+ */
+export async function tryReconcileOrphanPendingRootPoolMapping(
+  context: handlerContext,
+  rootChainId: number,
+  rootPoolAddress: string,
+): Promise<RootPool_LeafPool | null> {
+  const pending = await context.PendingRootPoolMapping.get(
+    PendingRootPoolMappingId(rootChainId, rootPoolAddress),
+  );
+  if (!pending) {
+    return null;
+  }
+
+  const matchingAggregators =
+    (await context.LiquidityPoolAggregator.getWhere({
+      rootPoolMatchingHash: { _eq: pending.rootPoolMatchingHash },
+    })) ?? [];
+  if (matchingAggregators.length !== 1) {
+    return null;
+  }
+
+  const leafAggregator = matchingAggregators[0];
+  const leafChainId = leafAggregator.chainId;
+  const leafPoolAddress = leafAggregator.poolAddress;
+
+  const rootPoolLeafPool: RootPool_LeafPool = {
+    id: RootPoolLeafPoolId(
+      rootChainId,
+      leafChainId,
+      rootPoolAddress,
+      leafPoolAddress,
+    ),
+    rootChainId,
+    rootPoolAddress,
+    leafChainId,
+    leafPoolAddress,
+  };
+  context.RootPool_LeafPool.set(rootPoolLeafPool);
+  context.PendingRootPoolMapping.deleteUnsafe(pending.id);
+
+  return rootPoolLeafPool;
+}
+
+/**
  * Attempts to load pool data, and if not found, checks if it's a RootCLPool
  * and loads the corresponding leaf pool data instead.
  *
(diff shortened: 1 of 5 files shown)


FonsSicca:
Closing in favor of a simpler skip-list.

## Why

On-chain bytecode inspection shows both orphan root pool addresses are **empty "always-revert" stubs**, not real pools:

- OP `0x333030A736B47D20346d82A473680658ac1C2b88` → bytecode `0x60806040525f80fdfe…` (Solidity 0.8.25, revert-only)
- Base `0x97Cd4EB683E29695DC93eec95d47d4E7a35E2112` → bytecode `0x6080604052600080fdfe…` (Solidity 0.8.19, revert-only)

Every standard pool selector (`token0()`, `token1()`, `tickSpacing()`, `factory()`, `fee()`, `stable()`, `chainid()`, `isRootPool()`, `getRatio()`) reverts. These are placeholder/sink contracts registered via `RootPoolCreated` and `GaugeCreated` for administrative purposes, but never wired to a real leaf pool.

The reactive reconciliation in this PR would look for a matching leaf `LiquidityPoolAggregator` via `rootPoolMatchingHash` — but no real leaf exists for either address, so the reconciliation is a no-op in practice. It's also speculative: we have no evidence of a real orphan that would benefit from this retry.

## Replacement approach

Follow-up PR adds an explicit skip-list for these two known-dead root pool addresses. `loadPoolDataOrRootCLPool` and `resolveLeafPoolForRootGauge` early-return silently when the pool is in the set — no reconciliation attempt, no noisy warn/error logs. Decision is documented in code so future readers understand these addresses are intentional sinks.

Leaves a cleaner codebase than the defensive-reconciliation approach: no import-cycle risk with `CrossChainPendingResolution`, no new helper surface, no entity writes in the hot path.

PR #611, opened 2026-04-17
fix(nfpm): scope position lookups by NFPM address to fix orphaned positions (#603)

### LINKED ISSUE #603: Diagnose and fix NFPM mint correlation — 443 orphan CL positions
## Parent PRD

#597

## What to build

During the audit window, 953 `NonFungiblePosition … not found` errors were emitted across 443 unique tokenIds, split between chains 10 (Optimism) and 5330 (Superseed). Every error is a downstream consequence of a position that was never created:

| op | chain 10 | chain 5330 |
|---|---|---|
| transfer | 293 | 252 |
| decreaseLiquidity | 188 | 192 |
| increaseLiquidity | 1 | 27 |

Concrete impact: tokenId 3689 on Superseed has 8 decrease + 7 increase events all missed — that user's TVL, fees, and staking contributions are invisible to every aggregator.

The producer-side warning `No CLPoolMintEvent found for NFPM.Transfer(mint)` fires only **5 times** in the same window, so the 443 affected positions cannot all be explained by simple correlation misses. Possible additional causes:

- **Cross-chain tokenId collision**: `findPositionByTokenId` in `NFPMCommonLogic.ts` calls `context.NonFungiblePosition.getWhere({ tokenId })` then filters by chainId in memory. If a tokenId exists on two chains and the wrong one leaks through or gets filtered out unexpectedly, lookups silently return empty.
- **Preload vs processing race**: handlers run twice in Envio; a scratchpad `CLPoolMintEvent` could be written in preload and deleted before processing reads it.
- **Downstream delete/burn cascades**: if a `NonFungiblePosition` is deleted on burn and a late event arrives, it appears orphaned.

Steps:
- Pick one affected tokenId per chain (e.g. 3689 on 5330, one from chain 10). Trace the event stream end-to-end.
- Determine the dominant root cause. Document which of the 3 hypotheses (or a new one) accounts for the bulk of the 443.
- Implement the fix at the right layer. A widened `getWhere` is explicitly NOT acceptable — fix the cause, not the symptom.
- Add tests covering cross-chain tokenId collision and any additional failure mode discovered.

HITL because investigation is required before the fix shape is known.

## Acceptance criteria

- [ ] Root-cause diagnosis posted as a comment (with example tokenIds and log/tx references).
- [ ] Fix implemented at the causal layer, not via widening queries.
- [ ] Test covering cross-chain tokenId collision.
- [ ] Test covering whichever additional failure mode was found.
- [ ] Post-fix backfill: `NonFungiblePosition not found` error count drops to near zero.

## Blocked by

None - can start immediately.

## User stories addressed

- User story 7
- User story 8

FonsSicca:
## Summary

Fixes #603 — the 443 orphan `NonFungiblePosition … not found` errors (953 events) observed on Optimism and Superseed during the Apr 8–17 audit. Root cause is a lookup-scope bug, not a correlation race.

## Problem

`findPositionByTokenId` filtered candidates by `(tokenId, chainId)`. But tokenIds are only unique within a single NFPM contract's `_nextId` counter — not across a chain. Optimism runs **two** NFPM contracts (`0xbB5D…AdF4` and `0x416b…F29`), so both counters emit overlapping tokenIds.

Trace:

1. NFPM-A mints `tokenId=N` → `position_A` created (`{10}-{poolX}-{N}`).
2. NFPM-B mints `tokenId=N` → `handleMintTransfer` calls `findPositionByTokenId(N, 10)`, which returns `[position_A]`.
3. The `existingPositions.length > 0` guard triggers and **silently returns** — `position_B` is never created, no warning logged.
4. Every downstream `Transfer`/`IncreaseLiquidity`/`DecreaseLiquidity` for NFPM-B's `tokenId=N` either misattributes to `position_A` or, once `position_A` is burned, logs "not found". This is why only 5 "No CLPoolMintEvent found" warnings explain 953 downstream errors — the producer-side correlation was working; the consumer-side lookup was wrong.

The PRD explicitly rules out widening `getWhere` as a fix — the lookup scope itself is what's broken.

## Fix

Scope the position identity by NFPM contract, not just chain:

- Add `nfpmAddress: String! @index` to `NonFungiblePosition` (and its snapshot), populated from `event.srcAddress` at mint time.
- Update `findPositionByTokenId(tokenId, chainId, nfpmAddress, context)` to filter by all three in memory (getWhere still supports only one indexed field).
- Thread `event.srcAddress` through every call site: `processNFPMTransfer`, `processNFPMIncreaseLiquidity`, `processNFPMDecreaseLiquidity`, and `createPositionFromCLPoolMint`.

Cross-chain collision (same tokenId on two chains, both with single NFPM) was already handled by the `chainId` filter and remains covered; the new `nfpmAddress` filter composes with it cleanly.

## Tests

- `NFPMCommonLogic.test.ts` — new cases for cross-chain collision, intra-chain multi-NFPM collision, and unknown-NFPM lookup returning empty.
- `NFPMTransferLogic.test.ts` — end-to-end regression: when NFPM-A already owns `(chainId=10, tokenId=540)`, NFPM-B's mint now correctly creates its own position on `poolY` without touching NFPM-A's.
- All existing NFPM/ALM/snapshot tests updated to carry `nfpmAddress` on their mock positions.

`pnpm qa --write` clean, `pnpm test` green (1026 passing, 32 skipped).

## Schema migration note

`NonFungiblePosition` gains a required field. Existing production data lacks it, so the indexer will re-sync positions from events on next deploy (standard Envio behavior for non-nullable schema additions). No backfill script required — the mint correlation path now populates `nfpmAddress` from `event.srcAddress`.

- **Primary signal** — `uerror` count for `NonFungiblePosition … not found during (transfer|increase liquidity|decrease liquidity)` should drop to near zero. Baseline: 953 errors / 443 unique tokenIds over the 9-day audit window. Target: < 10 errors over an equivalent window.
- **Log query** — grep `uerror` for `NonFungiblePosition with tokenId .* not found` and `\[handleRegularTransfer\] No positions provided`.
- **Producer-side invariant** — `No CLPoolMintEvent found for NFPM.Transfer(mint)` should stay at its current baseline (~5/9d). This fix is downstream-only and should not move the producer count.
- **Positive signal** — aggregate `NonFungiblePosition` entity count on Optimism should increase by the number of previously-skipped NFPM-B mints (rough order: 400+ over the audit window).
- **Failure signal / rollback** — if new mints fail to create positions (look for `tokenId=… nfpmAddress=… not found` after a successful `CLPool.Mint`), revert this PR. No operational mitigation needed beyond revert; the old behavior is strictly worse but not catastrophic.
- **Validation window** — 24h after deploy, compare 24h rolling count of the target error classes against the same window pre-deploy.
- **Owner** — indexer maintainers (per PRD #597).

Ties to user stories 7 & 8 in #597.

---

[Compound Engineering v2.62.0](https://github.com/EveryInc/compound-engineering-plugin)
🤖 Generated with Claude Opus 4.7 (1M context, extended thinking) via [Claude Code](https://claude.com/claude-code)

## Release Notes

* **New Features**
  * Added support for distinguishing non-fungible positions across multiple NFPM contracts on the same chain by including contract address in position identification.

* **Bug Fixes**
  * Fixed potential token ID collision detection when the same token ID exists across different NFPM contracts on the same chain.

### DIFF
--- a/schema.graphql
+++ b/schema.graphql
@@ -331,6 +331,7 @@ type NonFungiblePosition {
   id: ID! # Format: {chainId}-{poolAddress}-{tokenId} (NonFungiblePositionId)
   chainId: Int! # Chain ID where the NFT exists
   tokenId: BigInt! @index # Token ID of the NFT (from Transfer event)
+  nfpmAddress: String! @index # Address of the NFPM contract that emitted the Transfer; disambiguates intra-chain tokenId collisions across multiple NFPMs (e.g. Optimism has two)
   owner: String! @index # Checksum address of the current owner (from Transfer event)
   pool: String! @index # Address of the CL pool this position belongs to
   tickUpper: BigInt! @config(precision: 24) # Upper tick of the position range
@@ -350,6 +351,7 @@ type NonFungiblePositionSnapshot {
   id: ID! # Format: {chainId}-{tokenId}-{epochMs} (NonFungiblePositionSnapshotId)
   chainId: Int!
   tokenId: BigInt! @index
+  nfpmAddress: String! @index
   owner: String! @index
   pool: String! @index
   tickLower: BigInt! @config(precision: 24)
--- a/src/EventHandlers/NFPM/NFPMCommonLogic.ts
+++ b/src/EventHandlers/NFPM/NFPMCommonLogic.ts
@@ -21,18 +21,24 @@ export enum LiquidityChangeType {
 }
 
 /**
- * Finds a NonFungiblePosition entity by tokenId, filtering by chainId to avoid cross-chain collisions.
- * Uses a single-field getWhere (tokenId) then filters by chainId in memory because Envio's getWhere
- * supports only one filter field per call.
+ * Finds a NonFungiblePosition entity by tokenId, scoped to a specific chain AND NFPM contract.
+ *
+ * Uses a single-field getWhere (tokenId) then narrows in memory because Envio's getWhere supports
+ * only one filter field per call. Filtering by (chainId, nfpmAddress) is required because tokenIds
+ * are only unique within a single NFPM contract's counter — cross-chain and intra-chain (multiple
+ * NFPMs on the same chain, e.g. Optimism) collisions would otherwise leak the wrong position and
+ * silently misattribute liquidity, ownership, and stake state.
  *
  * @param tokenId - The token ID to search for
  * @param chainId - The chain ID to filter by
+ * @param nfpmAddress - The NFPM contract address (event.srcAddress) to filter by
  * @param context - The handler context for database operations
- * @returns Array of matching positions (should be 0 or 1), filtered by chainId
+ * @returns Array of matching positions (should be 0 or 1)
  */
 export async function findPositionByTokenId(
   tokenId: bigint,
   chainId: number,
+  nfpmAddress: string,
   context: handlerContext,
 ): Promise<NonFungiblePosition[]> {
   const positions = await context.NonFungiblePosition.getWhere({
@@ -43,9 +49,9 @@ export async function findPositionByTokenId(
     return [];
   }
 
-  // Filter by chainId to ensure we get the position from the correct chain
   return positions.filter(
-    (pos: NonFungiblePosition) => pos.chainId === chainId,
+    (pos: NonFungiblePosition) =>
+      pos.chainId === chainId && pos.nfpmAddress === nfpmAddress,
   );
 }
 
--- a/src/EventHandlers/NFPM/NFPMDecreaseLiquidityLogic.ts
+++ b/src/EventHandlers/NFPM/NFPMDecreaseLiquidityLogic.ts
@@ -52,10 +52,12 @@ export async function processNFPMDecreaseLiquidity(
 ): Promise<void> {
   // Get position by tokenId
   // Transfer should have already run and updated the placeholder, so position should exist when a DecreaseLiquidity event is processed
-  // Filter by chainId to avoid cross-chain collisions (same tokenId can exist on different chains)
+  // Filter by chainId AND nfpmAddress to avoid collisions: same tokenId can exist on different chains,
+  // and on chains with multiple NFPM contracts (e.g. Optimism) each NFPM has its own tokenId counter.
   const positions = await findPositionByTokenId(
     event.params.tokenId,
     event.chainId,
+    event.srcAddress,
     context,
   );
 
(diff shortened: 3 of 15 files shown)


FonsSicca:
@coderabbitai review

WooSungD (approved):

FonsSicca:
@coderabbitai review

FonsSicca:
## CodeRabbit review — addressed

Thanks for the review. Splitting the response by finding:

### Outside-diff comments (deferred to follow-up issues)

> **🔴 Critical — `src/EventHandlers/NFPM/NFPMTransferLogic.ts` (lines 41-50)**
> Include `nfpmAddress` in the `NonFungiblePosition` entity ID.

Real concern, deferred to **#621** (atomic schema migration: entity ID + delete `findPositionByTokenId` + replace 5 callers with direct `context.NonFungiblePosition.get()`).

In production today this collision is prevented by Velodrome's 1:1 NFPM↔CLFactory pairing — each NFPM mints into a disjoint pool set, so `(chainId, pool, tokenId)` is naturally unique. The current PR's read-side scoping is sufficient to fix the reported errors in #603. The structural fix (#621) makes the natural key explicit, kills the last `getWhere` scan in NFPM code (perf win), and guards against any future change to the factory architecture.

> **🟠 Major — `src/Snapshots/NonFungiblePositionSnapshot.ts` (lines 25-35)**
> Snapshot ID collision risk: include `nfpmAddress` or `poolAddress` to disambiguate positions across different NFPMs.

Real bug — already happens across the two Optimism NFPMs (different pools, same tokenId, same epoch). Deferred to **#620** because it ships independently of the entity-ID change (the position entity already carries `nfpmAddress` after this PR).

Both follow-ups are tracked under parent PRD **#618**.

### Nitpicks (addressed in 6ccf04d)

> **`test/EventHandlers/Pool/common.ts` (lines 39-41)** — Expose `nfpmAddress` from `setupCommon()` return value.

Addressed: exported a top-level `NFPM_ADDRESS` constant from `test/EventHandlers/Pool/common.ts` and surfaced it through `setupCommon()`'s return as well, so both call patterns work.

> **`test/EventHandlers/ALM/DeployFactoryV1.test.ts` (lines 36-38)** — Consider consolidating the repeated `nfpmAddress` literal.

Addressed: replaced the duplicated `toChecksumAddress("0xbB5...AdF4")` literal across 8 test files (NFPM/ALM/Aggregator) with the shared `NFPM_ADDRESS` constant. Future address changes are now a one-line edit.

> **`test/EventHandlers/NFPM/NFPMDecreaseLiquidityLogic.test.ts` (lines 45-60)** — Consider asserting `nfpmAddress` in the snapshot expectation.

Addressed: added `nfpmAddress: mockPosition.nfpmAddress` to the `expectSnapshotSet` matcher, so a regression where `nfpmAddress` fails to propagate from entity → snapshot is now caught.

`pnpm qa --write` clean, `pnpm test` green (1026 passing, 32 skipped).

Issue #675, opened 2026-05-04
bug(token-fetcher): `getTokenDetails` failure mode persists tokens with `decimals=0` and/or empty `symbol` instead of erroring/retrying


FonsSicca:
## Summary

\`getTokenDetails\` (\`src/Effects/fetchers/Token.ts\`) appears to have a silent-failure mode: when one or more of the multicall'd RPC reads (\`name()\`, \`symbol()\`, \`decimals()\`) fails, the result is persisted with default/empty values rather than retried or surfaced as an error. This is the suspected root cause for two visible symptoms in the audited dataset:

1. **Tokens persisted with \`decimals=0\`** — produces precision-overflow in swap-USD calculations (root cause for Pattern 1, see #668)
2. **Tokens persisted with empty \`symbol\`** (renders as \`?\`) — visible cosmetic damage on dashboards, plus suggests the same RPC failure may also corrupt other fields

## Reproduction

\`\`\`graphql
# Tokens with decimals=0 (should be ~6, ~8, or ~18 for legitimate tokens)
query {
  Token(where: { decimals: { _eq: \"0\" } }) {
    id chainId address symbol pricePerUSDNew
  }
}

# Tokens with empty/missing symbol
query {
  Token(where: { symbol: { _eq: \"\" } }) {
    id chainId address decimals pricePerUSDNew
  }
}
\`\`\`

The audit found 30+ tokens matching the first query and several matching the second.

## Likely root cause

\`getTokenDetails\` likely calls a multicall returning \`{ name, symbol, decimals }\`. If the underlying contract doesn't respond to one of those calls (e.g. some tokens have non-standard ABI for \`name\`), the multicall library typically returns \`undefined\` for that field and \`0\` / \`\"\"\` for the rest. The handler then persists those defaults without checking.

## Investigation pointers

- \`src/Effects/fetchers/Token.ts\` — \`getTokenDetails\` implementation
- \`src/PriceOracle.ts:21–48\` (\`createTokenEntity\`) — writes \`decimals: BigInt(tokenDetails.decimals)\` directly with no validation

## Recommended fix

In \`getTokenDetails\`:

1. If \`decimals\` returns \`0\` or fails to decode, **retry** with individual calls (single-shot per field) before persisting
2. If \`decimals\` is still \`0\` after retry AND on-chain bytecode at the token address is non-empty, log a warning and persist; if bytecode is empty, the address isn't a token contract and should be filtered earlier
[...]

PR #677, opened 2026-05-04
fix(token): preserve 0-decimals + stop overwriting metadata in refreshTokenPrice

FonsSicca:
## Summary

Two narrow, surgical fixes for issue #675:

1. **`fetchTokenDetails`**: replace `Number(decimalsResult) || 18` with `Number.isFinite`-guarded coercion so a contract that legitimately returns 0 (Base IDRX `0x649a...42cC`, FAT `0x5579...8E18`) is preserved instead of being silently rewritten to 18.

2. **`refreshTokenPrice`**: stop re-fetching `getTokenDetails` and stop overwriting `decimals` on the row. ERC20 metadata is immutable; the metadata refresh was the propagation mechanism that turned a single transient RPC failure during historical sync into permanent, sticky DB corruption (e.g. Soneium WETH `0x4200...0006` ending up with `decimals: 0n` despite a healthy RPC contract today).

## Why this works after a cache reset

The metadata-overwrite path is what was carrying poisoned cache values from `getTokenDetails` into Token rows on every refresh. With (2) removed, `getTokenDetails` is only ever called at *initial* token creation — all four creation paths (`Voter.WhitelistToken`, `SuperchainLeafVoter.WhitelistToken`, `processVotingRewardClaimRewards`, `createTokenEntity`) gate on `Token.get` returning undefined, so each token gets exactly one fetch attempt to populate metadata. Cache state no longer affects already-created rows.

A full effect-cache wipe + reindex now produces the correct outcome for the vast majority of tokens: every token's first fetch goes to RPC fresh, retried via `executeRpcWithFallback`'s primary + public-RPC path, and the result determines the row content permanently.

## Known gaps still present (intentionally not addressed here)

- **Non-contract addresses still create Token rows.** OP `0x3544773551EB240C2d8B499A62CAb994E7BB2662`, Lisk `0x18470019bF0E94611f15852F7e93cf5D65BC34CA`, Celo `0xa6438B910Fd62D07D85B1BAe4B6BfbedCf04d479` have no bytecode; their RPC reads always fail and the static fallback writes a row with empty symbol. The four `Token.set` call sites would need a bytecode-gate to prevent this — out of scope for this PR.
- **First-fetch failures for real tokens.** A small percentage of tokens whose initial `getTokenDetails` fetch fails during reindex will still have a bad row written from the fallback. Stays bad until the next cache reset; nothing post-creation re-fetches metadata.

## Test plan

- [x] `vitest run` — 1130 tests pass, 32 skipped (pre-existing). Includes:
  - Two new `fetchTokenDetails` tests (preserves 0; falls back to 18 on NaN).
  - Updated effect-call counts in `test/PriceOracle.test.ts` to reflect the dropped `getTokenDetails` call.
- [x] `tsc --noEmit` — clean.
- [x] `pnpm qa --write` — clean.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -103,3 +103,17 @@ Tests live in `test/` mirroring the `src/` structure. Uses Vitest with threaded 
 ## Lint/Format
 
 Uses Biome (not Prettier/ESLint). Space indentation, double quotes, organized imports. Run `biome check` before committing.
+
+## Agent skills
+
+### Issue tracker
+
+Issues live as GitHub issues at `velodrome-finance/indexer`; use the `gh` CLI. See `docs/agents/issue-tracker.md`.
+
+### Triage labels
+
+Canonical labels: `needs-triage`, `needs-info`, `ready-for-agent`, `ready-for-human`, `wontfix`. See `docs/agents/triage-labels.md`.
+
+### Domain docs
+
+Single-context layout: `CONTEXT.md` and `docs/adr/` at the repo root. See `docs/agents/domain.md`.
--- a/docs/agents/domain.md
+++ b/docs/agents/domain.md
@@ -1,0 +1,35 @@
+# Domain Docs
+
+How the engineering skills should consume this repo's domain documentation when exploring the codebase.
+
+This repo uses a **single-context** layout: one `CONTEXT.md` and one `docs/adr/` directory at the repo root.
+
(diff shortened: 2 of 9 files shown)


FonsSicca:
@coderabbitai review

FonsSicca:
@coderabbitai review

WooSungD (approved):

Issue #672, opened 2026-05-04
bug(soneium): 8 oUSDT/CL pools and tokens with missing symbols (`?`) show inflated USD totals — chain-specific token metadata + pricing failure


FonsSicca:
## Summary

Eight live pools on Soneium (chain \`1868\`) show inflated \`totalVolumeUSD\` and/or impossible \`totalFeesGeneratedUSD/totalVolumeUSD\` ratios. A common thread: many of these pools have a counterparty token whose \`symbol\` field is empty (renders as \`?\`) AND/OR whose \`pricePerUSDNew\` is implausibly high. This concentration on a single chain points at chain-specific configuration: either the Soneium price connectors in \`src/constants/price_connectors.json\` are misconfigured, or \`getTokenDetails\` fails systematically for some Soneium tokens.

This issue is filed separately from Patterns 1-4 because the chain-level concentration is the primary diagnostic signal.

## Affected pools

| pool | type | tokens (curr price) | totalVolumeUSD | totalFeesGeneratedUSD | swaps | symptom |
|---|---|---|---|---|---|---|
| \`1868-0xe424d1…2BEFA0466AAbEF82CB401534339ff2E838\` | CL | oUSDT (\$1/wl) / ? (\$1/wl) | \$1.09T | \$92.84M | 266,130 | AVG=\$4.1M/swap |
| \`1868-0xDaC26c…3ffe00ad0656f8Cd8752186d2f7D9ae789\` | CL | oUSDT (\$1/wl) / ? (\$2.32K) | \$698.53B | \$330.48M | 316,479 | AVG=\$2.2M/swap |
| \`1868-0xebb7B9…a2C1E930A761a2B41f4046971556e334d4\` | v2 | oUSDT (\$1/wl) / ? (\$1/wl) | \$52.66B | \$13.49M | 6,127 | AVG=\$8.6M/swap |
| \`1868-0x092B9A…C9DB9fA3F2D096329Df5E24404f13D836b\` | v2 | ? (\$2.32K) / HBI (\$1) | \$40 | \$14.83K | 904 | FEE_RATE=37,083% |
| \`1868-0x1EF23b…2b40A2312CD53cb2fC865e817CAB77E4D0\` | v2 | ? (\$0) / ? (\$0) | \$520 | \$27 | 5,156 | FEE_RATE=5% |
[...]

PR #678, opened 2026-05-04
fix(price-oracle): blacklist + canonical rebind for known-bad oracle reads (#669)

### LINKED ISSUE #669: bug(price-oracle): historical bad oracle reads contaminate lifetime totals on 8 live pools (SQUID, $Manatee, rsETH)
## Summary

Eight live pools across Optimism, Ink, and Swell have `totalVolumeUSD` and `totalFeesGeneratedUSD` warped by historical on-chain `SpotPriceAggregator` returns that captured wrong prices for one token at specific blocks. Once captured, the bad reads were summed into lifetime totals and cannot be unwound without reindex + a sanity gate.

The contamination has two flavors:

- **(a) Token currently mispriced** — \`Token.pricePerUSDNew\` is currently wrong (e.g. rsETH stored at \$3.62M, real value ~\$3.6K), so every read against this token (TVL, swap-USD) is warped both historically and now.
- **(b) Token historical spike, now reset** — the on-chain oracle returned a wrong rate at one or more historical blocks; subsequent refreshes returned \$0. Lifetime swap totals captured the spike; current TVL is fine because price is now \$0.

### (a) Currently mispriced token

| pool | chain | tokens (curr price) | totalVolumeUSD | totalFeesGeneratedUSD | swaps |
|---|---|---|---|---|---|
| \`1923-0xd91A99…D9950c8375575ebEcF2486\` | Swell | wstETH (\$0/wl) / **rsETH (\$3.62M/wl)** | \$69.02T | \$3.55B | 1,281 |
| \`1923-0xFC9E8c…56124106806a45eb8a4565b1E5C1\` | Swell | WETH (\$2.33K/wl) / **rsETH (\$3.62M/wl)** | \$2.10K | \$31.75B | 137 |
| \`10-0x482959…fdBD003edc785d19f16dECf\` | OP | tBTC (\$0/wl) / **\$Manatee (\$172.56K)** | \$26.15T | \$15.26B | 477 |
| \`10-0xD01b3e…4F31b164f4aC01f36B61B1C0DBfb\` | OP | **\$Manatee (\$172.56K)** / alUSD (\$0/wl) | \$50.03T | \$33.59B | 472 |

rsETH should be ~\$3,600 (based on ETH price), not \$3,620,000 — a ~1000× current-price error. \$Manatee priced at \$172K is implausible for any meme token.

### (b) Historical spike, current price now \$0

| pool | chain | tokens | totalVolumeUSD | totalFeesGeneratedUSD | swaps |
|---|---|---|---|---|---|
| \`57073-0x6eC366…6BDBA86bCDc321dCBCb6458a36\` | Ink | **SQUID (\$0)** / WETH (\$2.34K/wl) | \$13.17B | \$22.28M | 810 |
| \`57073-0x84D928…cBdCcb46808667F28A26bE2B6F4d1e0466\` | Ink | **SQUID (\$0)** / XVELO (\$0/wl) | \$51.47B | \$72.53M | 1,512 |
| \`57073-0x3Ac231…6f69d1CA125052D0d5f8356EE1606D13C4\` | Ink | **SQUID (\$0)** / PURPLE (\$0) | \$5.87B | \$8.37M | 943 |
| \`252-0x78b070…ce5f972E84EF9CBf3\` (overlap with Pattern 1) | Fraxtal | XVELO (\$0/wl) / frxETH (\$2.33K/wl) | \$348M | \$514K | 257 |

## Why this happens

Per-swap USD computation in \`PoolSwapLogic.ts\` / \`CLPoolSwapLogic.ts\` calls the on-chain \`SpotPriceAggregator.getManyRatesWithCustomConnectors\` directly at swap time and accumulates the result into \`totalVolumeUSD\` / \`totalFeesGeneratedUSD\`. There is no sanity check on the returned rate. If the oracle returns a bad rate at one of N swaps, the lifetime total is permanently contaminated.

Independent of swap-time accumulation, \`Token.pricePerUSDNew\` is refreshed hourly via \`refreshTokenPrice\` in \`src/PriceOracle.ts:64–187\`. Bad refreshes also persist because the oracle return is accepted without sanity checks beyond the \"\$0 means use last known price for 7 days\" fallback.

## Recommended fix (forward-only, requires fresh reindex)

Two-layer sanity gate, no schema change:

**Layer 1 — \`refreshTokenPrice\` (\`src/PriceOracle.ts\`):**
- Reject the new price if (a) it's >100× the previous non-zero price within one hour, (b) it's >\$1M per token unit and the token isn't whitelisted, or (c) one of the routing connectors used for this token is itself stuck at \$0
- On reject, keep the previous price and log

**Layer 2 — swap-USD calc (\`PoolSwapLogic.ts\`, \`CLPoolSwapLogic.ts\`):**
- After computing \`swapAmountUSD\`, if implied per-token-unit USD exceeds a threshold (e.g. \$1M/unit for non-WL, \$1B/unit for any), contribute 0 to \`totalVolumeUSD\` / \`totalFeesGeneratedUSD\`. Token-amount totals accumulate normally.
- Optionally increment a \`suspectSwapCount\` counter on \`LiquidityPoolAggregator\` for observability.

## Acceptance

- [ ] Add Layer-1 sanity gate in \`refreshTokenPrice\`
- [ ] Add Layer-2 sanity gate in v2 + CL swap handlers
- [ ] Verify on a fresh reindex that the 8 listed pools produce sane totals
- [ ] Verify rsETH on Swell refreshes to ~\$3K instead of \$3.6M

## Out of scope

- Pattern 1 (Fraxtal precision overflow) — separate issue, different root cause (decimals=0)
- Pattern 3 (fee-accumulator divergence) — separate issue, fees-exceed-volume independent of price
- Pattern 5 (Soneium oUSDT family) — overlaps but is chain-specific, separate issue

FonsSicca:
## Summary

Closes the four #669 cases where Velodrome's on-chain `SpotPriceAggregator` is structurally unusable for a given token. Instead of trying to filter bad reads heuristically, we bypass the oracle for those specific (chain, token) pairs.

Two override mechanisms in `src/PriceOverrides.ts`:

- **Blacklist** — force `pricePerUSDNew = 0n` and skip the oracle entirely. For tokens with no real off-chain market and an oracle that either (a) routes through a contaminated pool or (b) returns 0 except during transient swap-time spikes that contaminate lifetime totals.
- **Canonical rebind** — copy `pricePerUSDNew` verbatim from another already-priced `Token` entity. For tokens that are economically 1:1 with a token we price cleanly on a different chain (bridged superchain tokens, cross-chain LST/LRT duals).

Both fire **before** the oracle is consulted in `refreshTokenPrice`. If a rebind source has not yet been priced, we explicitly write 0 rather than falling back to the corrupt oracle — that fallback is the bug we are fixing.

## Tokens covered

| token | chain | mode | source / target |
|---|---|---|---|
| `$Manatee` `0x7909Bda5…` | Optimism (10) | blacklist | n/a — no real off-chain market |
| `SQUID` `0x2e3b8289…` | Ink (57073) | blacklist | n/a — oracle returns 0 except during transient swap spikes |
| `XVELO` `0x7f9AdFbd…` | Fraxtal, Lisk, Mode, Metal, Swell, Soneium, Celo, Unichain, Ink, Superseed | rebind | VELO/Optimism (`0x9560e827…`) — 1:1 SuperERC20 mint/burn via Hyperlane |
| `rsETH/Swell` `0xc3eaCf06…` | Swell (1923) | rebind | wrsETH/Base (`0xEDfa2360…`) — both Kelp wrappers of the same underlying restaked-ETH position; Aerodrome's wrsETH/WETH pool gives a clean Base price |

## Why not Llama / GeckoTerminal / Dexscreener

Investigated all three as fallbacks. None is sufficient on its own:

- $Manatee has no off-chain coverage at all — every external API reads the same broken DEX pool we already read on-chain.
- SQUID coverage is sparse and lifetime totals can't be reconstructed retroactively.
- XVELO has spotty coverage and the canonical 1:1 relationship to VELO is more authoritative than a third-party feed.
- rsETH/Swell is already covered by our own Base index via the deep Aerodrome wrsETH/WETH pool, so an external API would be a strict downgrade.

## What's NOT in this PR

- Spike / per-unit-cap heuristics. An earlier version of this PR added a Layer-1 spike gate (`isAcceptablePrice`); on review it added complexity without addressing any case the explicit overrides don't, and could false-positive on legitimate new-listing pumps. Reverted in `c5aa4aa`. If a future bad token appears, the right lever is to extend `PriceOverrides.ts`.
- Issue #672 (Soneium-specific symptoms). Distinct root causes — `getTokenDetails` failing silently and possible Soneium connector misconfiguration. Same override mechanism could be reused for any Soneium tokens identified later, but #672 needs its own investigation and is explicitly out of scope here.

## Test plan

- [x] `pnpm test` — 1148 passed, 32 skipped
- [x] `tsc --noEmit` clean
- [x] `pnpm qa --write` clean
- [ ] Reindex one of the affected pools (e.g. rsETH/WETH on Swell) and confirm lifetime totals are sane
- [ ] Confirm rsETH on Swell refreshes to ~\$2.4K (matching Base wrsETH), \$Manatee and SQUID stay at \$0, XVELO refreshes to VELO/OP price

## Summary by CodeRabbit

* **New Features**
  * Added token price override mechanisms: blacklist tokens to force zero price and bypass oracle fetching, and configure rebind rules to inherit prices from canonical source tokens across chains.

### DIFF
--- a/src/PriceOracle.ts
+++ b/src/PriceOracle.ts
@@ -12,6 +12,7 @@ import {
   roundBlockToInterval,
 } from "./Effects/Index";
 import { EffectType, rpcGateway } from "./Effects/RpcGateway";
+import { getRebindTarget, isBlacklistedToken } from "./PriceOverrides";
 import { setTokenPriceSnapshot } from "./Snapshots/TokenPriceSnapshot";
 export interface TokenPriceData {
   pricePerUSDNew: bigint;
@@ -85,6 +86,44 @@ export async function refreshTokenPrice(
 
   if (!shouldRefresh) {
     return token;
+  }
+
+  // Issue #669: blacklist + canonical rebind override the oracle for known-bad
+  // (chain, token) pairs. See src/PriceOverrides.ts for rationale per token.
+  if (isBlacklistedToken(chainId, token.address)) {
+    const updated: Token = {
+      ...token,
+      pricePerUSDNew: 0n,
+      lastUpdatedTimestamp: new Date(blockTimestampMs),
+    };
+    context.Token.set(updated);
+    return updated;
+  }
+
+  const rebindTarget = getRebindTarget(chainId, token.address);
+  if (rebindTarget) {
+    const sourceToken = await context.Token.get(
+      TokenId(rebindTarget.chainId, rebindTarget.address),
+    );
+    const sourcePrice = sourceToken?.pricePerUSDNew ?? 0n;
+    const updated: Token = {
+      ...token,
+      pricePerUSDNew: sourcePrice,
+      lastUpdatedTimestamp: new Date(blockTimestampMs),
+    };
+    context.Token.set(updated);
+    if (sourcePrice > 0n) {
+      setTokenPriceSnapshot(
+        token.address,
+        chainId,
+        blockNumber,
+        new Date(blockTimestampMs),
+        sourcePrice,
+        token.isWhitelisted,
+        context,
+      );
+    }
+    return updated;
   }
 
   try {
--- a/src/PriceOverrides.ts
+++ b/src/PriceOverrides.ts
@@ -1,0 +1,120 @@
+import { TokenId, toChecksumAddress } from "./Constants";
+
+export interface RebindTarget {
+  readonly chainId: number;
+  readonly address: string;
+}
+
+/**
+ * Tokens whose on-chain price oracle is structurally unusable: every read either
+ * returns a value derived from a broken DEX pool (no real off-chain market exists
+ * to disagree) or is silent except for transient swap-time spikes that contaminate
+ * lifetime totals. For these, the only correct price is 0 — pool-USD calcs already
+ * route through the counterparty when one side is $0.
+ *
+ * Issue #669: $Manatee (no real market; tBTC/$Manatee pool routes price through
+ * tBTC and produces $76K-$92K; alETH/$Manatee disagrees with alUSD/$Manatee by 10x).
+ * Issue #669: SQUID (oracle returns 0 at every stable read; the $13B contaminated
+ * total comes from transient single-block spikes during swaps).
+ */
+const BLACKLIST: ReadonlySet<string> = new Set([
+  TokenId(10, toChecksumAddress("0x7909Bda52eAf7C3cc12745E727Eb527a485241D8")), // $Manatee / Optimism
+  TokenId(
+    57073,
+    toChecksumAddress("0x2e3b82891d1B2b90655597110cCA9b6587607e0c"),
+  ), // SQUID / Ink
+]);
+
+/**
+ * Chains where Velodrome's SuperERC20 deployments live at the same address as
+ * their canonical Optimism counterpart (1:1 mint/burn through Hyperlane).
+ */
+const SUPERCHAINS = [
+  252, // Fraxtal
+  1135, // Lisk
+  34443, // Mode
+  1750, // Metal
+  1923, // Swell
+  1868, // Soneium
+  42220, // Celo
+  130, // Unichain
+  57073, // Ink
+  5330, // Superseed
+] as const;
+
+/**
+ * Tokens whose price should be copied verbatim from another (already-priced)
+ * Token entity, bypassing the on-chain oracle entirely. Each entry declares a
+ * canonical `source` and every `target` (chain + address) that mirrors it.
+ *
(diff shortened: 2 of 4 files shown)


FonsSicca:
@coderabbitai review

WooSungD (approved):

PR #679, opened 2026-05-04
feat(aggregator): warn on fee/volume USD divergence (#670)

FonsSicca:
## Summary

Adds a soft invariant warning in `updateLiquidityPoolAggregator` that fires when a pool's running `totalFeesGeneratedUSD` crosses 5% of `totalVolumeUSD`. Real swap fee tiers cap at ~1%, so this threshold gives plenty of headroom while still catching the kind of divergence reported in #670 (50% – 14M%+ fee-vs-volume ratios on 6+ live pools).

The warning is tagged `[FEE_VOLUME_DIVERGENCE]` (matching the existing `[STAKED_TICK_DRIFT]` style) and only emits on the *crossing* event, so logs stay clean once a pool has tripped.

Also adds regression tests asserting fee USD ≤ 1% of volume USD for both V2 (`processPoolFees`) and CL (`processCLPoolSwap`) at realistic fee tiers (1, 5, 30, 100 bps). These pass against the current code, locking in the contract so future changes can't silently widen the divergence without test failures.

Closes acceptance criteria #2 (regression test) and #3 (aggregator-level invariant log) of #670. Root-cause investigation and reindex verification (#1, #4) are out of scope for this PR — the warning is the visibility hook needed to catch them.

## Test plan

- [x] `pnpm test` — full suite passes (1156 / 1188, 32 skipped pre-existing)
- [x] `tsc --noEmit` clean
- [x] `pnpm qa --write` clean
- [ ] Reindex affected pools and verify the warning surfaces for the known offenders (Lisk USDT/SYGX, Mode WETH/CARTEL, Ink kBTC/SolvBTC)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #670, opened 2026-05-04
bug(swap-fee): totalFeesGeneratedUSD exceeds totalVolumeUSD by 50%–14M% on 6 live pools — fee path diverged from volume path


FonsSicca:
## Summary

Six live and several stale pools have \`totalFeesGeneratedUSD / totalVolumeUSD\` ratios between **50%** and **14,338,520%**. Real swap fee tiers cap at ~1%, so fees this much higher than volume cannot be a price problem — they indicate the **fee accumulator code path computes USD value differently than the volume accumulator** in some swap branches.

This is independent of the price-oracle bugs (Pattern 2): even tokens with currently-correct prices (kBTC \$79K, SolvBTC \$464K) produce impossible fee/volume ratios.

## Affected pools (live, last touched ≤30d)

| pool | chain | type | tokens | totalVolumeUSD | totalFeesGeneratedUSD | fee rate | swaps |
|---|---|---|---|---|---|---|---|
| \`57073-0x039fC1…99C8F975ab0a0a8f7a68Dbf8C18ba03d0\` | Ink | v2 | kBTC (\$79K/wl) / SolvBTC (\$464K/wl) | \$20.53K | \$48.19M | **234,711%** | 520 |
| \`57073-0x3170b9…355F1057F457FEdF4c8074946659Dc92D2\` | Ink | CL | kBTC (\$79K/wl) / SolvBTC (\$464K/wl) | \$112.75K | \$2.40M | **2,131%** | 291 |
| \`1868-0xCD80Af…2043e233A56700A32e75312aaaa13677D8\` | Soneium | CL | ? (\$2.32K) / wstETH (\$2.92K/wl) | \$645.30K | \$7.61M | **1,180%** | 39,264 |
| \`57073-0x402BD7…9D1927256561FaD8d80F04Ba20C5F3b3D3\` | Ink | v2 | WETH (\$2.34K/wl) / SolvBTC (\$464K/wl) | \$323 | \$1.74K | **537%** | 429 |
| \`57073-0xaF54c9…f078686Ed15ba274C86eA03AdAA43AA31f\` | Ink | v2 | XVELO (\$0/wl) / MYST (\$0) | \$38.53K | \$19.34K | **50%** | 3,820 |
[...]

Issue #666, opened 2026-05-04
bug(cl-staked): negative stakedReserve0/1 running counters on 166 CL pools across all chains


FonsSicca:
## Summary

166 CL pools across every Velodrome chain have `stakedReserve0` or `stakedReserve1` running counters that have drifted **negative** over their lifetime. The fields are typed `BigInt @config(precision: 76)` and represent staked reserves in raw token units — they should never go below zero.

The defensive clamp in `src/Aggregators/LiquidityPoolAggregator.ts:444–457` (`stakedReserve0 > 0n ? stakedReserve0 : 0n`) masks this at the USD layer, so `currentLiquidityStakedUSD` reads as $0 (not negative), but the raw `stakedReserve0/1` fields persist as negative and are queryable.

## Distribution

Counts of pools where `stakedReserve0 < 0` OR `stakedReserve1 < 0` (deployed indexer at `https://indexer.us.hyperindex.xyz/3d71402/v1/graphql`):

| chainId | chain      | pools |
|---------|------------|-------|
| 10      | Optimism   | 69    |
| 1868    | Soneium    | 18    |
| 57073   | Ink        | 15    |
| 1135    | Lisk       | 13    |
| 1923    | Swell      | 11    |
| 130     | Unichain   | 8     |
| 5330    | Superseed  | 8     |
| 252     | Fraxtal    | 6     |
| 1750    | Metal      | 6     |
| 34443   | Mode       | 6     |
| 42220   | Celo       | 6     |
| **total** | | **166** |

All affected pools are CL (`isCL: true`).

## Example pools

```
chain=10 pool=10-0xCAfFfE7… cl=true sR0=-252490000 sR1=2727052158459901178 swaps=1758
chain=10 pool=10-0xA61FBA4… cl=true sR0=127708019961736589974376 sR1=-135152857691846445422805 swaps=2597
chain=10 pool=10-0x8E55FF6… cl=true sR0=69892965020 sR1=-372720456187214354447837 swaps=6715
chain=10 pool=10-0x20efb6B… cl=true sR0=-205923431041 sR1=624408760215215304960574 swaps=4825
chain=10 pool=10-0xE96CEd7… cl=true sR0=-1132972076516482452154 sR1=1038943274066645804743 swaps=301
chain=10 pool=10-0x8Fc5014… cl=true sR0=-4350338725843752391483 sR1=4343863089208118709421 swaps=2099
[...]

PR #680, opened 2026-05-04
feat(aggregator): warn on reserve0/reserve1 negative drift (#674)

### LINKED ISSUE #674: bug(superseed): negative `reserve0`/`reserve1` (unstaked) on Superseed pools — distinct from #666 staked-side drift
## Summary

Multiple pools on Superseed (chain \`5330\`) have negative \`reserve0\` and/or \`reserve1\` running counters. These are the **unstaked** reserve fields, distinct from the \`stakedReserve0\`/\`stakedReserve1\` drift tracked in #666 (which affects staked CL pools across all chains).

The negative-unstaked pattern appears concentrated on Superseed, suggesting a chain-specific cause: either Superseed's pool factory configuration, gauge wiring, or some pool/handler interaction is producing reserve deltas that don't cancel out.

## Reproduction

\`\`\`graphql
query {
  LiquidityPoolAggregator(
    where: {
      chainId: { _eq: 5330 }
      _or: [
        { reserve0: { _lt: \"0\" } }
        { reserve1: { _lt: \"0\" } }
      ]
    }
  ) {
    id chainId isCL
    reserve0 reserve1
    totalLiquidityUSD
    numberOfSwaps
    lastUpdatedTimestamp
  }
}
\`\`\`

(Run against the deployed indexer at \`https://indexer.us.hyperindex.xyz/3d71402/v1/graphql\`.)

## Why this is distinct from #666

- #666: \`stakedReserve0\`/\`stakedReserve1\` (the *staked* portion of CL pool reserves) negative on 166 CL pools across all chains. Root cause is in the staked tick-crossing accumulator path.
- This issue: \`reserve0\`/\`reserve1\` (the unstaked / total pool reserves) negative on Superseed pools specifically. Different code path — \`PoolSyncLogic.ts\` for v2 or the equivalent CL reserve tracking.

## Likely root causes (to investigate)

1. **Superseed-specific event ordering**: Superseed may have a sequencer behavior that emits Sync/Swap/Mint/Burn in an order the indexer doesn't expect
2. **Mint/Burn missing on Superseed**: dynamic-contract registration may have raced and missed initial Mint events, leading to Burns subtracting from a 0 baseline
3. **Initial-state seeding**: the pool was registered after liquidity was already present; seed reserves were 0 and Burns drove them negative

## Acceptance

- [ ] Identify which Superseed pool/handler combination produces the drift
- [ ] Add a regression test
- [ ] Verify a fresh reindex produces non-negative reserves on Superseed
- [ ] If the same pattern shows up on a different chain, expand scope to multi-chain

## Out of scope

- #666 (negative \`stakedReserve0/1\` on 166 CL pools) — different accumulator
- Pattern 4 (Lisk ION/WETH inflated TVL) — different (positive-side drift)

FonsSicca:
## Summary

Adds a soft invariant warning in `updateLiquidityPoolAggregator` that fires when a pool's `reserve0` or `reserve1` crosses from non-negative to negative. Reserves track LP-deposited capital and should never go below zero — issue #674 reports negative `reserve0`/`reserve1` running counters concentrated on Superseed (chain `5330`).

The warning is tagged `[NEGATIVE_RESERVE_DRIFT]` (matching the existing `[STAKED_TICK_DRIFT]` and `[FEE_VOLUME_DIVERGENCE]` style introduced in #679) and only emits on the **crossing** event, so logs stay clean once a pool has tripped. The warning includes pool address, chainId, prior value, applied delta, and resulting negative value — enough to pinpoint which (chain, pool, handler) combination produced the drift.

The check is chain-agnostic. While #674 is concentrated on Superseed, the issue notes "If the same pattern shows up on a different chain, expand scope to multi-chain" — emitting the warning everywhere means no extra change is needed if it does.

Adds 4 regression tests covering crossing on each reserve side, no-warn when staying non-negative, and no-warn when already negative (since the warning only fires on the transition, not while the pool stays in a corrupted state).

Closes acceptance criterion #2 (regression test) of #674. Root-cause identification (#1) and reindex verification (#3) are out of scope for this PR — the warning is the visibility hook needed to catch them, the same staged approach used by #679 for issue #670.

## What's NOT in this PR

- Clamping reserves to 0 on negative drift. The team's pattern (#666 acceptance: "log a warning so future drifts are detected"; #679 likewise warns without clamping) preserves the raw value so it's queryable for diagnostics.
- Root-cause investigation of the Superseed-specific source. Likely candidates noted in #674 (event ordering, dynamic-contract registration race, initial-state seeding) require running the indexer with the new warning to identify which handler trips it.

## Test plan

- [x] `pnpm test` — 1137 passed, 32 skipped (matches pre-PR baseline)
- [x] `tsc --noEmit` clean
- [x] `pnpm qa --write` clean
- [ ] Reindex Superseed pools and confirm `[NEGATIVE_RESERVE_DRIFT]` warnings surface for the affected pools, then use the captured (pool, delta) info to root-cause the producer

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reserve validation handling to emit diagnostic warnings instead of halting operations when reserves drift negative.

* **Tests**
  * Added regression test coverage for reserve drift warning scenarios across snapshot epoch boundaries.

### DIFF
--- a/src/Aggregators/LiquidityPoolAggregator.ts
+++ b/src/Aggregators/LiquidityPoolAggregator.ts
@@ -499,6 +499,22 @@ export async function updateLiquidityPoolAggregator(
       );
     }
 
+    // Soft invariant (issue #674): reserve0/reserve1 track LP-deposited capital
+    // and should never go below zero. Logged at each snapshot epoch boundary
+    // while still negative (≤1/hour per pool) so a persistent drift stays
+    // visible in recent logs without flooding them and without aborting the
+    // indexer or mutating state.
+    if (updated.reserve0 < 0n) {
+      context.log.warn(
+        `[NEGATIVE_RESERVE_DRIFT][updateLiquidityPoolAggregator] Pool ${current.poolAddress} on chain ${current.chainId} reserve0 is negative (${updated.reserve0}). Reserves are LP-deposited capital and should never go below zero.`,
+      );
+    }
+    if (updated.reserve1 < 0n) {
+      context.log.warn(
+        `[NEGATIVE_RESERVE_DRIFT][updateLiquidityPoolAggregator] Pool ${current.poolAddress} on chain ${current.chainId} reserve1 is negative (${updated.reserve1}). Reserves are LP-deposited capital and should never go below zero.`,
+      );
+    }
+
     // Only set lastSnapshotTimestamp when we actually created a snapshot (epoch boundary)
     updated = {
       ...updated,
--- a/test/Aggregators/LiquidityPoolAggregator.test.ts
+++ b/test/Aggregators/LiquidityPoolAggregator.test.ts
@@ -477,6 +477,120 @@ describe("LiquidityPoolAggregator Functions", () => {
         expect(negativeStakedReserveWarnings().length).toBe(0);
       });
     });
+
+    // Regression test for issue #674: pool reserves should never go negative.
+    // The aggregator emits a warning under [NEGATIVE_RESERVE_DRIFT] at each
+    // snapshot epoch boundary while still negative (≤1/hour per pool) so a
+    // persistent drift stays visible in recent logs without flooding them and
+    // without aborting the indexer. Concentrated on Superseed (chain 5330)
+    // but the warning is chain-agnostic.
+    describe("negative reserve drift warning (issue #674)", () => {
+      const previousEpoch = () =>
+        new Date(timestamp.getTime() - 2 * 60 * 60 * 1000);
+
+      const negativeReserveWarnings = () => {
+        const warnMock = vi.mocked(mockContext.log?.warn);
+        const calls = warnMock?.mock.calls ?? [];
+        return calls.filter((args) =>
+          String(args[0] ?? "").includes("[NEGATIVE_RESERVE_DRIFT]"),
+        );
+      };
+
+      it("warns at snapshot epoch when reserve0 is negative", async () => {
+        await updateLiquidityPoolAggregator(
+          { incrementalReserve0: -100n },
+          {
+            ...(liquidityPoolAggregator as LiquidityPoolAggregator),
+            reserve0: 50n,
+            reserve1: 1000n,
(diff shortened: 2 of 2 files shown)


src/Aggregators/LiquidityPoolAggregator.ts
```diff
@@ -424,6 +424,23 @@ export async function updateLiquidityPoolAggregator(
...
+  // without flooding them and without aborting the indexer or mutating state.
+  if (updated.reserve0 < 0n && current.reserve0 >= 0n) {
+    context.log.warn(
+      `[NEGATIVE_RESERVE_DRIFT][updateLiquidityPoolAggregator] Pool ${current.poolAddress} on chain ${current.chainId} reserve0 crossed negative: ${current.reserve0} + ${diff.incrementalReserve0 ?? 0n} = ${updated.reserve0}. Reserves are LP-deposited capital and should never go below zero.`,
+    );
+  }
+  if (updated.reserve1 < 0n && current.reserve1 >= 0n) {
+    context.log.warn(
+      `[NEGATIVE_RESERVE_DRIFT][updateLiquidityPoolAggregator] Pool ${current.poolAddress} on chain ${current.chainId} reserve1 crossed negative: ${current.reserve1} + ${diff.incrementalReserve1 ?? 0n} = ${updated.reserve1}. Reserves are LP-deposited capital and should never go below zero.`,
+    );
```
WooSungD: should there also be a warning when `reserve0` and `reserve1` are both negative?

FonsSicca: in that case both warnings are fired

WooSungD: ah I misread - all good.

these warnings would also fire only once right? similar to the previous practice of warnings only firing once and never again, should a similar change be made?

FonsSicca: you're right, it's the same problem. will do a commit for it

WooSungD (approved):

PR #681, opened 2026-05-04
fix(cl-staked): warn on stakedReserve0/1 negative drift (#666)

FonsSicca:
## Summary

- Closes part of #666 (observability) — adds a `[NEGATIVE_STAKED_RESERVE_DRIFT]` `context.log.warn` in `updateLiquidityPoolAggregator` that fires only on the **crossing event** (non-negative → negative), so future drifts surface in logs without flooding them and without aborting the indexer or mutating state.
- Mirrors the [NEGATIVE_RESERVE_DRIFT] pattern from #674/#680 for the unstaked-reserves sibling bug.
- Keeps the defensive `max(0, _)` clamp at the USD-conversion site in place — root-cause fix (sign error in tick crossings vs. stake/unstake ordering vs. sparse-list invariant drops) is a separate follow-up. See acceptance criteria #1, #3, #4 of the issue.

## Why crossing-only

166 affected pools already have negative running counters; warning on every subsequent update during a reindex would flood logs. Logging only the transition gives one warn-line per drift event, which is enough to bisect.

## Test plan

- [x] `vitest run test/Aggregators/LiquidityPoolAggregator.test.ts` — 4 new tests in `describe(\"negative staked reserve drift warning (issue #666)\")` cover both axes (sR0, sR1) and both no-warn cases (stays non-negative; already negative)
- [x] Full aggregator file: PASS (35) FAIL (0)
- [x] `tsc --noEmit` clean
- [x] `pnpm qa --write` clean (no fixes needed)
- [ ] Watch staging reindex logs for `[NEGATIVE_STAKED_RESERVE_DRIFT]` to bisect which event handler / chain / pool emits the first crossing — feeds into the root-cause follow-up

### DIFF
--- a/src/Aggregators/LiquidityPoolAggregator.ts
+++ b/src/Aggregators/LiquidityPoolAggregator.ts
@@ -424,32 +424,6 @@ export async function updateLiquidityPoolAggregator(
     lastUpdatedTimestamp: timestamp,
   };
 
-  // Soft invariant (issue #666): stakedReserve0/stakedReserve1 are running
-  // counters of staked LP-deposited capital and should never go negative. 166
-  // CL pools across all chains have drifted negative; the defensive
-  // max(0, _) clamp below at the USD-conversion site masks the symptom but
-  // the raw fields persist as negative. Log under
-  // [NEGATIVE_STAKED_RESERVE_DRIFT] only on the crossing event
-  // (non-negative → negative) so future drifts surface in logs without
-  // flooding them and without aborting the indexer or mutating state.
-  // Mirrors the [NEGATIVE_RESERVE_DRIFT] pattern from #674.
-  if (
-    (updated.stakedReserve0 ?? 0n) < 0n &&
-    (current.stakedReserve0 ?? 0n) >= 0n
-  ) {
-    context.log.warn(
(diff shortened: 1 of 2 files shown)


src/Aggregators/LiquidityPoolAggregator.ts
```diff
@@ -424,6 +424,32 @@ export async function updateLiquidityPoolAggregator(
...
+      `[NEGATIVE_STAKED_RESERVE_DRIFT][updateLiquidityPoolAggregator] Pool ${current.poolAddress} on chain ${current.chainId} stakedReserve0 crossed negative: ${current.stakedReserve0 ?? 0n} + ${diff.incrementalStakedReserve0 ?? 0n} = ${updated.stakedReserve0 ?? 0n}. Staked reserves are LP-deposited capital and should never go below zero.`,
+    );
+  }
+  if (
+    (updated.stakedReserve1 ?? 0n) < 0n &&
+    (current.stakedReserve1 ?? 0n) >= 0n
+  ) {
+    context.log.warn(
+      `[NEGATIVE_STAKED_RESERVE_DRIFT][updateLiquidityPoolAggregator] Pool ${current.poolAddress} on chain ${current.chainId} stakedReserve1 crossed negative: ${current.stakedReserve1 ?? 0n} + ${diff.incrementalStakedReserve1 ?? 0n} = ${updated.stakedReserve1 ?? 0n}. Staked reserves are LP-deposited capital and should never go below zero.`,
+    );
```
WooSungD: similar comment as on #680 - should a warning be fired for cases when both `stakedReserve0` and `stakedReserve1` are negative

FonsSicca: same response - in that case both warnings are fired

WooSungD: same comment as [here](https://github.com/velodrome-finance/indexer/pull/680#discussion_r3207806522)

WooSungD (approved):

PR #683, opened 2026-05-04
feat(price-oracle): cross-chain RPC prefetch when rebind source is unpriced

FonsSicca:
## Summary

- When a token is rebound to another chain's source token (e.g. XVELO on superchains → VELO/Optimism, rsETH/Swell → wrsETH/Base) and the local `Token` entity for that source has not yet been priced, the rebind branch now reaches across to the source chain's oracle directly instead of writing $0.
- New `src/ChainBlockTime.ts` exports `estimateBlockAtTimestamp(chainId, timestampSec)` with per-chain `(block, timestamp, blocktime)` anchors for the only chains that act as rebind sources today: Optimism and Base.
- The estimated source-chain block is rounded to a 1-hour bucket via the existing `roundBlockToInterval`, so the prefetch fills the **same** effect-cache slot the source chain's own indexer will hit on its eventual native refresh — zero extra RPCs once the source catches up.

## Why

Background, in case the rebind path is unfamiliar (see #678): some target tokens have a structurally broken on-chain oracle. Their price is rebound to a canonical source token on another chain (e.g. every superchain's XVELO copies VELO/Optimism's price, since SuperERC20 + Hyperlane preserve 1:1 parity). Until #678, if the source `Token` entity hadn't been priced yet the rebind silently wrote `0` and waited for the source chain's indexer to catch up. That's correct but wasteful — and on cold-sync it produces a window of $0 prices.

This PR closes that gap. Multi-chain handlers run independently, so the target chain's handler may run ahead of the source chain's. Now, when the local source is unpriced, we estimate the source chain's block from the target's wall-clock timestamp and dispatch `getTokenPrice` against the source chain. The cache key is `(tokenAddress, chainId, blockNumber)`, so when the source chain's own indexer eventually reaches the same hour bucket, it hits the cache slot we already filled.

## Worked example (also in `test/PriceOracle.test.ts`)

```
Fraxtal handler at unix T = 1_700_100_000 wants the price of XVELO,
which rebinds to VELO/Optimism. Local Token entity for VELO/OP unpriced.

  estimateBlockAtTimestamp(10, T)              → 112_250_611
  roundBlockToInterval(112_250_611, 10)        → 112_249_800   (1-hour bucket)
  context.effect(getTokenPrice, {
    tokenAddress: VELO_OP,
    chainId:      10,
    blockNumber:  112_249_800,
  })

When the OP indexer eventually refreshes VELO at any block in that
same hour, it rounds to 112_249_800 and hits the cache slot above.
```

## Behavior guarantees

- If the prefetch returns a non-zero price → use it, write a snapshot.
- If the prefetch also returns 0 → still write 0. We **never** fall through to the local (corrupt) oracle — that's the whole point of the rebind.
- If the source chain has no anchor configured (= no entry in `CHAIN_ANCHORS`) → skip the prefetch and write 0 as before.

## Files

- `src/ChainBlockTime.ts` — new
- `test/ChainBlockTime.test.ts` — new
- `src/PriceOracle.ts` — rebind branch updated
- `test/PriceOracle.test.ts` — rebind tests updated for prefetch behavior

## Test plan

- [x] `pnpm test` — full suite passes (1160 passed / 32 skipped)
- [x] `tsc --noEmit` clean
- [x] `pnpm qa` (biome) clean
- [ ] Reindex one cold-sync scenario locally and confirm the target chain's tokens populate non-zero prices ahead of the source chain's natural catch-up
- [ ] Confirm the source chain's native refresh hits the prefetched cache slot (no duplicate RPCs)

## Summary by CodeRabbit

* **New Features**
  * Added cross-chain block-time estimation functionality enabling calculation of approximate block numbers from UNIX timestamps
  * Enhanced price oracle with source-chain prefetching capability for rebind/canonical oracle override scenarios, improving pricing accuracy when local source tokens have unavailable pricing data

* **Tests**
  * Test coverage added for cross-chain block-time estimation and source-chain price oracle prefetch scenarios

### DIFF
--- a/src/ChainBlockTime.ts
+++ b/src/ChainBlockTime.ts
@@ -1,0 +1,59 @@
+/**
+ * Per-chain anchor pairs for converting a wall-clock timestamp into an
+ * approximate block number on that chain.
+ *
+ * Each anchor is a single (block, unix-timestamp) reference pair plus the
+ * chain's nominal block time. Estimation is plain arithmetic:
+ *
+ *     estimatedBlock = anchorBlock + floor((timestamp - anchorTimestamp) / blocktime)
+ *
+ * This is a HEURISTIC — real block times drift slightly, so the estimate can
+ * be off by a handful of blocks over months. Callers should pair this with
+ * `roundBlockToInterval` (1-hour buckets ≈ 1800 blocks on 2s chains), which
+ * absorbs drift and aligns the estimate with the cache key the source chain's
+ * own indexer will eventually use.
+ *
+ * Only chains that act as a `source` in `PriceOverrides.REBINDS` need an
+ * anchor — that is currently Optimism and Base.
+ *
+ * NOTE: `blocktimeSeconds` here must match the implicit blocktime in
+ * `roundBlockToInterval` (Effects/Token.ts), otherwise the prefetched cache
+ * slot won't line up with the source chain's eventual native refresh.
+ */
+interface ChainAnchor {
+  readonly block: number;
+  readonly timestamp: number; // unix seconds
+  readonly blocktimeSeconds: number;
+}
+
+const CHAIN_ANCHORS: ReadonlyMap<number, ChainAnchor> = new Map([
+  // Optimism Bedrock genesis: block 105_235_063 at 2023-06-06 16:28:23 UTC.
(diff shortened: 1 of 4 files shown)


FonsSicca:
@coderabbitai review

WooSungD (approved):

Issue #668, opened 2026-05-04
bug(fraxtal): swap-USD totals warped by transient sfrxETH oracle spikes — needs ratio-based price guard


FonsSicca:
## Update — original hypothesis disproved by post-#677 reindex

A fresh Fraxtal-only reindex on `main` (which includes #677) confirms the originally suspected `decimals=0` cause is **not** what's warping these totals — every Fraxtal token, including all five referenced in the issue (wstETH, ezETH, sfrxETH, frxETH, XVELO), is persisted with the correct decimals.

The actual cause is **transient absurd `pricePerUSDNew` reads from the Fraxtal V3 oracle**. Two tokens are affected: sfrxETH and FXB20291231. Their refresh streams contain isolated windows where the oracle returned values 15–30 orders of magnitude away from the surrounding baseline, then returned to the baseline. Each bad refresh that gets persisted poisons every swap priced against it for the duration of that hour bucket, and because cumulative volume/fees are append-only, **a few corrupted snapshots permanently dominate the lifetime aggregate**.

## Proposed fix — ratio-based price-spike rejection in `refreshTokenPrice`

```ts
const PRICE_SPIKE_RATIO_THRESHOLD = 10n;          // 1 OOM
const PRICE_SPIKE_STALENESS_MS = 7 * 24 * 60 * 60 * 1000;

const oldPrice = token.pricePerUSDNew;
const oldAgeMs = blockTimestampMs - (token.lastUpdatedTimestamp?.getTime() ?? 0);
const ratioExceeds =
  oldPrice > 0n && currentPrice > 0n &&
  (currentPrice > oldPrice * PRICE_SPIKE_RATIO_THRESHOLD ||
   oldPrice > currentPrice * PRICE_SPIKE_RATIO_THRESHOLD);

if (ratioExceeds && oldAgeMs < PRICE_SPIKE_STALENESS_MS) {
  context.log.warn(`[priceSpikeRejected] ${token.address} chain=${chainId} old=${oldPrice} new=${currentPrice}`);
  currentPrice = oldPrice;        // reject — keep last accepted
}
```

Rules:

- **First-fetch** (`oldPrice == 0n`) is exempt — accept as ground truth.
- **Stale anchor** (`oldAgeMs >= PRICE_SPIKE_STALENESS_MS`) is exempt — slow legitimate drift over weeks isn't permanently frozen.
- Threshold = **10× (1 OOM)** — see "Heuristic calibration" below.
[...]

PR #686, opened 2026-05-08
fix(helpers): cap swap-USD when token decimals are malformed (#668)

FonsSicca:
## Summary
- Five Fraxtal pools (chain `252`) had lifetime `totalVolumeUSD` / `totalFeesGeneratedUSD` warped into quadrillion-range — implied per-swap USD `1e19`–`1e25`, ~`1e15×` larger than physically possible. Root cause: at least one route token (likely `wstETH` / `ezETH` / `sfrxETH`) persisted with `decimals=0`, after which the swap-USD path normalized as if 18-decimal and inflated the result by `1e18`.
- Add a sanity cap inside `calculateTokenAmountUSD` (the single chokepoint for swap, fee, and reserve USD math): if the implied per-raw-unit USD (`pricePerUSDNew / 10^tokenDecimals`) exceeds **$1**, zero the contribution. No real ERC20 has a smallest unit worth more than $1, so legitimately priced tokens — including 0-decimal stablecoins like IDRX (~$0.00006/unit) — pass through untouched. Anything tripping the cap is malformed and would otherwise poison aggregates.

## Out of scope
- The `getTokenDetails` failure mode that lets `decimals=0` get persisted in the first place — separate issue.
- Pattern-2 price-spike contamination on other chains — separate issue.

## Test plan
- [x] New unit tests cover: 18-decimal token misread as 0 → zaps; legitimate `decimals=0` low-priced token → not zapped; normal 18-decimal swap → unchanged; one-corrupted-token swap routed through `calculateTotalUSD` keeps the healthy side's USD.
- [x] Full project test suite (`pnpm test`) — 1176 passing, 0 regressions.
- [x] `tsc --noEmit` clean, `pnpm qa --write` clean.
- [ ] Operational: verify a fresh reindex produces sane lifetime totals on the 5 affected Fraxtal pools.

### DIFF
--- a/src/Helpers.ts
+++ b/src/Helpers.ts
@@ -73,6 +73,14 @@ export function calculateTokenAmountUSD(
   tokenDecimals: number,
   pricePerUSDNew: bigint,
 ): bigint {
+  // Issue #668: reject malformed (decimals, price) tuples whose implied per-raw-unit
+  // USD exceeds $1 — no real ERC20 has a smallest unit worth more than $1, so this
+  // only triggers when a token's decimals were lost (typically persisted as 0
+  // instead of 18 after an RPC failure). One such token can otherwise warp
+  // lifetime swap-USD totals into the quadrillions.
+  if (pricePerUSDNew > TEN_TO_THE_18_BI * 10n ** BigInt(tokenDecimals)) {
+    return 0n;
+  }
   const normalizedAmount = normalizeTokenAmountTo1e18(amount, tokenDecimals);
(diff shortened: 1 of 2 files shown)


FonsSicca:
Closing — the upstream fix in #677 (stop overwriting immutable `decimals` on every `refreshTokenPrice`) addresses the root cause of #668. Verifying with a fresh Fraxtal reindex; this guard would only be defense-in-depth on the residual gap #677 acknowledges, with an arbitrary $1/raw-unit threshold that's hard to defend.

Issue #673, opened 2026-05-04
bug(emissions): `totalEmissionsUSD` is $0 on every pool, every chain


FonsSicca:
## Summary

\`LiquidityPoolAggregator.totalEmissionsUSD\` is \$0 on **every pool across every chain** in the deployed indexer (\`https://indexer.us.hyperindex.xyz/3d71402/v1/graphql\`). This is a real data gap — gauges are emitting rewards in token units (\`totalEmissions\`), but the USD-denominated counterpart is universally zero.

## Reproduction

\`\`\`graphql
query {
  LiquidityPoolAggregator(
    where: { totalEmissionsUSD: { _gt: \"0\" } }
    limit: 5
  ) { id chainId totalEmissions totalEmissionsUSD }
}
\`\`\`

Returns 0 results across all 11 Velodrome chains plus Aerodrome (Base).

\`\`\`graphql
query {
  LiquidityPoolAggregator(
    where: { totalEmissions: { _gt: \"0\" } }
    limit: 5
  ) { id chainId totalEmissions totalEmissionsUSD }
}
\`\`\`

Returns many results — \`totalEmissions\` is non-zero (in token units) but \`totalEmissionsUSD\` stays \$0.

## Likely root cause

Two candidates:

1. **USD conversion site missing**: emission events are tracked in \`totalEmissions\` (token units) but a corresponding \`* tokenPrice\` multiplication step into \`totalEmissionsUSD\` was either never written or is short-circuited.
2. **Wrong token used for USD price**: emissions are denominated in the gauge's reward token (typically VELO / xVELO / AERO depending on chain). If the USD conversion looks up the *pool's* token0/token1 instead of the *reward token*, and neither pool token matches the reward, the result silently rounds to 0.

## Investigation pointers

- \`src/EventHandlers/Gauges/GaugeSharedLogic.ts\` — emission accumulator
- \`src/EventHandlers/Voter/VoterCommonLogic.ts\` — also touches emissions
- \`src/Aggregators/LiquidityPoolAggregator.ts\` — search for \`totalEmissionsUSD\` writes
- \`src/Helpers.ts\` — \`calculateTokenAmountUSD\` should be called against the reward token

## Acceptance

- [ ] Identify where \`totalEmissionsUSD\` should be set
- [ ] Verify the conversion uses the reward token's price (VELO/xVELO/AERO), not pool token prices
[...]

PR #693, opened 2026-05-11
Fix #676: unstick $0-trapped tokens with a uniform 1h refresh throttle

### LINKED ISSUE #676: bug(price-oracle): 30-day backoff permanently traps tokens at $0 on transient oracle failures (OP connectors stuck 1,000+ days)
## Summary

The 30-day backoff in \`refreshTokenPrice\` (\`src/PriceOracle.ts:79–84\`) is intended to bound RPC waste on permanently unpriceable tokens. It has the side effect of **permanently trapping legit tokens** that hit \$0 due to a transient oracle issue (single bad block, RPC outage, connector-graph misconfiguration, etc.). Once a token's \`pricePerUSDNew\` is \$0 for >30 days, no further refresh attempts happen — the token stays at \$0 forever.

## Evidence

Optimism's main whitelisted tokens have \`Token.pricePerUSDNew = 0\` and \`lastUpdatedTimestamp\` more than **1,000 days stale** (i.e. ~3 years — corresponds to initial indexer deploy). The deployed indexer at \`https://indexer.us.hyperindex.xyz/3d71402/v1/graphql\` shows:

\`\`\`graphql
query {
  Token(where: {
    chainId: { _eq: 10 }
    isWhitelisted: { _eq: true }
    pricePerUSDNew: { _eq: \"0\" }
  }) { id symbol address pricePerUSDNew lastUpdatedTimestamp }
}
\`\`\`

Returns major tokens (WETH, USDC, USDT, OP, VELO, DAI) all stuck at \$0 with timestamps from 2022–2023. Live swaps continue to compute USD correctly because each \`Pool.Swap\` event calls the on-chain oracle directly — but the cached \`Token.pricePerUSDNew\` field, used by every TVL calculation across the indexer, is permanently zero.

## Why this matters

\`Token.pricePerUSDNew\` is the canonical source for:
- \`LiquidityPoolAggregator.totalLiquidityUSD\` (TVL)
- All snapshot USD fields downstream of the token entity
- Anything that doesn't go through the live swap path

Once a major token like WETH is stuck at \$0, every pool holding WETH on Optimism has TVL under-reported. This is why Optimism's chain-level TVL appears so much lower than expected on dashboards.

## Current workaround (technical debt to remove)

\`src/PriceOracle.ts:107–120\` and \`src/Constants.ts\` contain an \`AFFECTED_CHAINS\` set + cache-bypass branch:

\`\`\`ts
// TEMPORARY: Bypass effect cache for affected chains with \$0 cached results.
// These chains had broken oracle connectors that cached \$0 prices permanently.
// The rpcGateway effect (cache: false) re-fetches from now-fixed connectors.
// Remove after one full reindex with fixed connectors.
\`\`\`

This patches the symptom (stuck \$0 → re-fetch live) but doesn't address the underlying refresh-skip logic. The bypass is also chain-scoped, so any future chain hitting the same trap requires a code change.

## Likely root cause

\`src/PriceOracle.ts:79–84\`:

\`\`\`ts
const shouldRefresh =
  !token.lastUpdatedTimestamp ||
  (token.pricePerUSDNew === 0n
    ? blockTimestampMs - token.lastUpdatedTimestamp.getTime() < THIRTY_DAYS_MS
    : blockTimestampMs - token.lastUpdatedTimestamp.getTime() >=
      MS_IN_AN_HOUR);
\`\`\`

The intent (\"don't waste RPC on permanently dead tokens\") is reasonable, but the policy is too coarse — it doesn't distinguish whitelisted majors from random meme tokens.

## Recommended fix

Tiered backoff based on token importance:

1. **Whitelisted tokens** — always refresh hourly, never enter backoff. RPC cost is bounded by the small whitelist set; correctness matters more than cost for these.
2. **Connector tokens** (anything in \`src/constants/price_connectors.json\`) — same: always refresh hourly.
3. **Other tokens** — keep current 30-day backoff, but add an escape hatch: if a swap event for the token comes in, force a refresh attempt regardless of last-update age. This way a meme token that suddenly becomes active gets re-priced.

After this fix lands together with #669 (Layer-1 sanity gate at refresh) and a fresh reindex, the \`AFFECTED_CHAINS\` workaround in \`PriceOracle.ts\` and \`Constants.ts\` should be removed.

## Acceptance

- [ ] Whitelisted tokens on Optimism refresh hourly on fresh reindex (no more 1000-day-stale timestamps)
- [ ] Connector tokens refresh hourly regardless of price history
- [ ] \`AFFECTED_CHAINS\` cache-bypass branch removed from \`PriceOracle.ts\` and the set removed from \`Constants.ts\`
- [ ] Optimism chain TVL on dashboards reflects the actual on-chain value rather than a fraction of it

## Out of scope

- The on-chain oracle's correctness — upstream, not addressable from the indexer
- #669 (Layer-1 sanity gate at \`refreshTokenPrice\`) — separate but complementary; this issue is about *which* tokens get refreshed, that issue is about *what* values are accepted on refresh

FonsSicca:
Closes #676.

## Summary

The 30-day `$0` backoff in `refreshTokenPrice` was meant to bound RPC waste on permanently unpriceable tokens. In practice it trapped every Optimism major (WETH, USDC, USDT, OP, VELO, DAI) at `$0` with `lastUpdatedTimestamp` from 2023 — they hit the broken-oracle window at deploy, the 30-day clock expired before the connectors were fixed, and they were never refreshed again. `Token.pricePerUSDNew` feeds every TVL/snapshot path that doesn't go through the live swap oracle, so Optimism's chain-level TVL has been under-reported ever since.

## Approach

Rather than a tiered allow-list (whitelisted/connector/neither), this is a deeper simplification: **every `refreshTokenPrice` call is event-triggered**, so the function is in the wrong place to decide "is this token dead". The 30-day gate is dropped in favour of a uniform 1-hour throttle. RPC cost is bounded by two existing mechanisms:

1. **Envio's effect cache** — `getTokenPrice` is keyed on `(token, chainId, hourly-rounded block)`; calls within the same hour bucket hit the cache.
2. **The 1-hour throttle** at function entry, gated on `lastUpdatedTimestamp`.

`lastUpdatedTimestamp` is always bumped on attempt (even when the oracle returns `$0`), so the throttle advances. This subsumes the #673 `oracleDeployed`-conditioned preservation and the `AFFECTED_CHAINS` cache-bypass workaround — both become dead code.

Worst-case RPC: ~5.5 RPCs/sec across all chains if every one of the ~20k currently-trapped tokens fires every hour. The global limit is 20k/sec. Three orders of magnitude of headroom.

## Changes

- `src/PriceOracle.ts` — collapse `shouldRefresh` to the throttle; drop the `AFFECTED_CHAINS` bypass branch; drop the `oracleDeployed` timestamp-preservation; refresh the function's JSDoc.
- `src/Constants.ts` — delete `AFFECTED_CHAINS`.
- `test/PriceOracle.test.ts` — invert the buggy `should NOT retry after 30-day backoff` test; consolidate `#673`'s pre/post-oracle preservation tests into "always-advance" characterisations; delete the `rpcGateway bypass for affected chains (Change C)` describe block.
- `test/Aggregators/LiquidityPoolAggregator.test.ts` — update the `$0` token refresh test for the new throttle semantics.
- `test/EventHandlers/Voter/DistributeRewardEmissionsUSD.test.ts` — `#673` regression test no longer relies on the 30-day backoff to gate the refresh; uses the 1-hour throttle instead.

Net diff: +77 / -296.

## Acceptance criteria (from #676)

- [x] Whitelisted tokens on Optimism refresh hourly on fresh reindex *(satisfied by the general rule)*
- [x] Connector tokens refresh hourly regardless of price history *(satisfied by the general rule)*
- [x] `AFFECTED_CHAINS` cache-bypass branch removed from `PriceOracle.ts` and the set removed from `Constants.ts`
- [ ] Optimism chain TVL on dashboards reflects actual on-chain value *(needs human — requires a fresh reindex against the deployed indexer)*

## Test plan

- [x] `tsc --noEmit` clean
- [x] `pnpm qa --write` clean
- [x] `pnpm test` — 1195 passed / 32 skipped / 0 failed
- [ ] Deploy + reindex Optimism, confirm WETH (`0x4200…0006`) `pricePerUSDNew` is non-zero and `lastUpdatedTimestamp` is fresh *(needs human — requires deployed indexer)*
- [ ] Spot-check Optimism chain TVL on dashboards post-reindex *(needs human — requires deployed indexer)*

## Release Notes

* **Bug Fixes**
  * Standardized token price refresh throttling to uniform 1-hour intervals across all chains for more consistent behavior.
  * Improved oracle price recovery handling when prices are unavailable.
  * Simplified chain-specific price refresh logic for better reliability.

[Review Change Stack](https://app.coderabbit.ai/change-stack/velodrome-finance/indexer/pull/693)

### DIFF
--- a/src/Constants.ts
+++ b/src/Constants.ts
@@ -38,18 +38,6 @@ export const SECONDS_IN_A_DAY = BigInt(86400);
 export const SECONDS_IN_A_WEEK = BigInt(604800);
 export const SECONDS_IN_FOUR_YEARS = BigInt(126144000);
 
-/**
- * Chains with historically broken oracle connectors that cached $0 prices.
- * Used by refreshTokenPrice to bypass the effect cache on these chains.
- * TEMPORARY: Remove after one full reindex with fixed connectors.
- */
-export const AFFECTED_CHAINS: ReadonlySet<number> = new Set([
-  42220, // Celo
-  1923, // Swell
-  252, // Fraxtal
-  1868, // Soneium
-]);
-
 /** Snapshot interval: 1 hour in milliseconds (for epoch-aligned snapshots) */
 export const SNAPSHOT_INTERVAL_IN_MS = 60 * 60 * 1000;
 
--- a/src/PriceOracle.ts
+++ b/src/PriceOracle.ts
@@ -1,7 +1,6 @@
 import type { Token, handlerContext } from "generated";
 import { estimateBlockAtTimestamp } from "./ChainBlockTime";
 import {
-  AFFECTED_CHAINS,
   CHAIN_CONSTANTS,
   MS_IN_AN_HOUR,
   PriceOracleType,
@@ -12,7 +11,6 @@ import {
   getTokenPrice,
   roundBlockToInterval,
 } from "./Effects/Index";
-import { EffectType, rpcGateway } from "./Effects/RpcGateway";
 import { getRebindTarget, isBlacklistedToken } from "./PriceOverrides";
 import { setTokenPriceSnapshot } from "./Snapshots/TokenPriceSnapshot";
 export interface TokenPriceData {
@@ -50,18 +48,19 @@ export async function createTokenEntity(
 }
 
 /**
- * Refreshes a token's price data if the update interval has passed.
+ * Refreshes a token's price subject to a uniform 1-hour throttle, dispatching
+ * to blacklist, cross-chain rebind, or on-chain oracle paths as configured.
  *
- * This function checks if enough time has passed since the last update (1 hour),
- * and if so, fetches new price data for the token. The token entity is updated
- * in the database with the new price and timestamp.
+ * `lastUpdatedTimestamp` is bumped on every attempt — including those that
+ * return $0 — so the throttle advances and we don't re-hit the oracle on every
+ * subsequent event for stuck tokens.
  *
- * @param {Token} token - The token entity to refresh
- * @param {number} blockNumber - The block number to fetch price data from
- * @param {number} blockTimestamp - The timestamp of the block in seconds
- * @param {number} chainId - The chain ID where the token exists
- * @param {any} context - The database context for updating entities
- * @returns {Promise<Token>} The updated token entity
+ * @param token - The token entity to refresh.
+ * @param blockNumber - Block at which to fetch price (rounded to an hourly bucket for cache hits).
+ * @param blockTimestamp - Block timestamp in seconds.
+ * @param chainId - Chain the token lives on.
+ * @param context - Envio handler context.
+ * @returns The updated token entity (or the unchanged input when throttled).
  */
 export async function refreshTokenPrice(
   token: Token,
@@ -71,19 +70,14 @@ export async function refreshTokenPrice(
   context: handlerContext,
 ): Promise<Token> {
   const blockTimestampMs = blockTimestamp * 1000;
-  const THIRTY_DAYS_MS = 30 * 24 * 60 * 60 * 1000;
 
-  // Refresh logic:
-  // - Missing timestamp → always refresh
-  // - $0 price for <30 days → retry (connector fix or Change A may self-heal)
-  // - $0 price for >30 days → stop retrying (accepted as unpriceable, bounds RPC waste)
-  // - Non-zero price → refresh on hourly interval
+  // Issue #676: uniform 1-hour throttle for all tokens (regardless of current
+  // price). RPC cost is bounded by the throttle plus Envio's hourly-rounded
+  // effect cache key on `getTokenPrice`. Safe because we always bump
+  // `lastUpdatedTimestamp` below, so the throttle advances even on $0 results.
   const shouldRefresh =
     !token.lastUpdatedTimestamp ||
-    (token.pricePerUSDNew === 0n
-      ? blockTimestampMs - token.lastUpdatedTimestamp.getTime() < THIRTY_DAYS_MS
-      : blockTimestampMs - token.lastUpdatedTimestamp.getTime() >=
-        MS_IN_AN_HOUR);
+    blockTimestampMs - token.lastUpdatedTimestamp.getTime() >= MS_IN_AN_HOUR;
 
   if (!shouldRefresh) {
     return token;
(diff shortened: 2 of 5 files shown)


WooSungD (approved):

FonsSicca:
CodeRabbit triage:

- **`test/Aggregators/LiquidityPoolAggregator.test.ts:1244-1266` (deterministic clock)** — applied in `75976de`. Replaced `new Date()` / `Date.now()` with a fixed `2024-01-01T12:00:00Z` reference. 1195/1195 tests pass.

- **`src/PriceOracle.ts:189-211` (V3 fallback timestamp aliasing)** — declining for this PR. The V3 fallback branch already bumped `lastUpdatedTimestamp` on every retry before this PR (unchanged from `main` — only the `$0+$0` branch's conditional-preserve was removed by this PR, and that branch doesn't fire when there's a previous non-zero price to fall back to). The aliasing concern is real but pre-existing and orthogonal to the #676 fix. Proper fix needs a separate `lastSuccessfulPriceTimestamp` field on the `Token` schema + codegen + migration — heavy lift better tracked as its own issue. Will open one.

FonsSicca:
Addressing the two outside-diff findings from the CodeRabbit pass:

**1. `src/PriceOracle.ts:189-211` — V3 fallback advances both the throttle clock and the staleness clock.** Real bug, but the fix requires splitting `lastUpdatedTimestamp` into two fields on the `Token` schema (refresh throttle vs. last-successful-price timestamp), which is a codegen + snapshot-path + migration change that doesn't belong in #676's throttle PR. Tracking as #695.

**2. `test/Aggregators/LiquidityPoolAggregator.test.ts:1244-1266` — use a fixed clock.** Already addressed in the current branch: line 1244 reads `const now = new Date("2024-01-01T12:00:00Z");` and `blockTimestamp` at line 1266 is derived from `now.getTime()`. No further action.

FonsSicca:
Correction on my last comment: the two-clock fix is already tracked as #694, not #695 (which I just opened and closed as a duplicate — apologies, missed the existing issue when triaging).

PR #697, opened 2026-05-12
Plumb usedDefault flag through executeRpcWithFallback (closes #691)

### LINKED ISSUE #691: refactor(effects): plumb `usedDefault` flag through executeRpcWithFallback
## Parent

Spun out of [PR #690](https://github.com/velodrome-finance/indexer/pull/690) (bytecode-gate for Token row creation sites, issue #677).

## Context

`executeRpcWithFallback` returns the same value shape on success and on "both RPCs exhausted, here's the hardcoded fallback constant" (`true` for the bytecode gate, `TOKEN_DETAILS_FALLBACK` for `getTokenDetails`, etc.). The outer effects can't tell those cases apart, so they rely on fragile **shape-detection heuristics** to decide whether to cache:

```ts
// src/Effects/Token.ts:60
if (result.name === "" && result.symbol === "") {
  context.cache = false; // "looks like TOKEN_DETAILS_FALLBACK"
}

// src/Effects/Token.ts:106
if (result.pricePerUSDNew === 0n) {
  context.cache = false; // "looks like the \$0 fallback"
}
```

Failure modes of the heuristic:

- `getTokenPrice` false-positive: a token with genuinely no oracle path has its real \$0 cache-skipped on every run — wasteful but bounded.
- `hasContractBytecode` collision: the fail-open value is `true`, indistinguishable from a real contract. Caused the bug fixed in PR #690 by switching to \`cache: false\`, which lost cross-run amortization of positives.

## What to build

Replace the shape heuristics with a proper signal from the gateway helper.

1. Widen `executeRpcWithFallback`'s return type from \`T\` to \`{ value: T; usedDefault: boolean }\` (or similarly-named field). \`usedDefault: true\` only when the helper returned the hardcoded fallback constant after both primary and fallback RPCs were exhausted.
2. Thread the flag through each gateway handler's output schema (`RpcGatewayOutputByType`) and through `callRpcGateway`'s typed wrapper.
3. Update outer effects (`getTokenDetails`, `getTokenPrice`, `hasContractBytecode`) to set \`context.cache = false\` based on \`usedDefault\` instead of value shape. Remove the heuristic guards.
4. Re-enable \`cache: true\` on \`hasContractBytecode\` so positive bytecode results amortize across runs (one \`eth_getCode\` per address per cache lifetime, not per indexer run).
5. Update tests in \`test/Effects/RpcGateway.test.ts\` to assert \`usedDefault\` is set correctly on the fail-open path and unset on real RPC answers.

## Acceptance criteria

- [x] \`executeRpcWithFallback\` returns a structured result that carries the fallback-vs-real signal.
- [x] \`getTokenDetails\` and \`getTokenPrice\` no longer use value-shape heuristics to control caching.
- [x] \`hasContractBytecode\` has \`cache: true\` restored and uses the new flag to skip caching only on the fail-open path.
- [x] New unit tests cover: (a) real RPC success → \`usedDefault: false\`, cached; (b) primary fail + fallback success → \`usedDefault: false\`, cached; (c) both exhausted → \`usedDefault: true\`, not cached.
- [x] Existing tests for the 4 Token.set creation sites still pass.
- [x] \`pnpm qa --write\` and \`pnpm test\` clean.

All verified in PR #697.

## Blocked by

None — can start immediately after PR #690 merges.

FonsSicca:
## Summary
- `executeRpcWithFallback` now returns `{ value, usedDefault }`; `usedDefault: true` iff the caller's fallback constant was returned after primary + fallback RPCs were exhausted.
- `RpcGatewayOutputByType` and each handler's `outputSchema` carry `usedDefault: boolean` so the signal propagates through the gateway without recasting.
- `getTokenDetails`, `getTokenPrice`, and `hasContractBytecode` drop the value-shape heuristics (`name === "" && symbol === ""`, `pricePerUSDNew === 0n`, `hasCode === true`) and gate `context.cache = false` on `usedDefault` instead.
- `hasContractBytecode` regains `cache: true` so real positive bytecode reads amortise across runs again (one `eth_getCode` per address per cache lifetime, not per indexer run).
- Sidecar fix in `src/PriceOracle.ts`: `let currentPrice` instead of `const` — pre-existing typo from PR #689 that broke `tsc` and the three #668 spike-rejection tests on `main`. Included here because the issue requires `pnpm qa --write` and `pnpm test` clean.

## Acceptance criteria (from #691)
- [x] `executeRpcWithFallback` returns a structured result with the fallback-vs-real signal (`RpcResult<T>` in `src/Effects/RpcGateway.ts`).
- [x] `getTokenDetails` and `getTokenPrice` no longer use value-shape heuristics (`src/Effects/Token.ts`).
- [x] `hasContractBytecode` has `cache: true` restored and uses `usedDefault` to skip caching only on the fail-open path (`src/Effects/Bytecode.ts`).
- [x] New unit tests cover (a) real RPC success → `usedDefault:false`, (b) primary fail + fallback success → `usedDefault:false`, (c) both exhausted → `usedDefault:true` (`test/Effects/RpcGateway.test.ts` — `executeRpcWithFallback` describe block).
- [x] Existing tests for the 4 `Token.set` creation sites still pass.
- [x] `pnpm qa` clean.
- [x] `pnpm test` clean (1217 passed, 33 skipped, 0 failed).

## Test plan
- [x] `tsc --noEmit` — no errors.
- [x] `pnpm qa` — Biome clean across 261 files.
- [x] `pnpm test` — 100/100 test files, 1217/1217 tests passing.
- [x] New `test/Effects/Bytecode.test.ts` locks `hasContractBytecode` cache semantics (cache stays on for real reads, off only when `usedDefault:true`).
- [x] Updated assertions in `test/Effects/RpcGateway.test.ts` confirm `usedDefault` is `false` on real RPC answers and `true` only on fail-open paths for every operation.
- [ ] Live indexer run on a chain with a flaky fallback RPC, to observe that positive bytecode reads now cache across runs *(needs human — requires shared infra)*.

### DIFF
--- a/src/Effects/Bytecode.ts
+++ b/src/Effects/Bytecode.ts
@@ -9,12 +9,12 @@ export { fetchHasContractBytecode } from "./fetchers/Bytecode";
  * addresses (e.g. Lisk 0x1847…34CA) that would otherwise persist Token rows
  * with empty `symbol`/`name` from the static fallback in {@link getTokenDetails}.
  *
- * Caching is disabled because {@link handleHasContractBytecode} fails open
- * (returns `hasCode: true` on transient RPC failure), and a cached `true` would
- * defeat the gate for that address permanently. The Token row itself acts as
- * the natural cache: callers `context.Token.get` first and only invoke this
- * effect for addresses not yet persisted, so at most one `eth_getCode` per
- * new token address per indexer run.
+ * Caching is enabled so positive bytecode results amortise across runs (one
+ * `eth_getCode` per address per cache lifetime). The fail-open path — where
+ * {@link handleHasContractBytecode} returns `hasCode: true` after both primary
+ * and fallback RPCs are exhausted — is detected via the gateway's `usedDefault`
+ * flag (issue #691) and skips caching so a transient RPC failure can be retried
+ * on the next run instead of being pinned to `true` until the cache evicts.
  *
  * @param input.address - Address to query.
  * @param input.chainId - Chain ID for RPC client.
@@ -31,7 +31,7 @@ export const hasContractBytecode = createEffect(
       hasCode: S.boolean,
     },
     rateLimit: false,
-    cache: false,
+    cache: true,
   },
   async ({ input, context }) => {
     const result = await callRpcGateway(context, {
@@ -39,6 +39,10 @@ export const hasContractBytecode = createEffect(
       address: input.address,
       chainId: input.chainId,
     });
+
+    if (result.usedDefault) {
+      context.cache = false;
+    }
 
     return { hasCode: result.hasCode };
   },
--- a/src/Effects/RpcGateway.ts
+++ b/src/Effects/RpcGateway.ts
@@ -56,6 +56,7 @@ const RPC_GATEWAY_OPERATIONS = {
       name: S.string,
       decimals: S.number,
       symbol: S.string,
+      usedDefault: S.boolean,
     },
   },
   [EffectType.GET_TOKEN_PRICE]: {
@@ -68,6 +69,7 @@ const RPC_GATEWAY_OPERATIONS = {
     outputSchema: {
       pricePerUSDNew: S.bigint,
       priceOracleType: S.string,
+      usedDefault: S.boolean,
     },
   },
   [EffectType.GET_TOKENS_DEPOSITED]: {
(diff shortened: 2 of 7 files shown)


WooSungD (approved):

FonsSicca:
@coderabbitai review

PR #715, opened 2026-05-14
fix(swap): pick min-USD leg to resist scam-token volume inflation (#699)

### LINKED ISSUE #699: Volume formula: pick trusted leg, not `token0 || token1`
## What to build

The CL and V2 swap handlers compute per-swap `volumeUSD` with:

```ts
const volumeInUSD = token0UsdValue !== 0n ? token0UsdValue : token1UsdValue;
```

This picks token0's USD leg whenever token0 has any non-zero price, falling back to token1 only when token0's price is `0n`. When one side of the pair has a corrupted `Token.pricePerUSDNew` (e.g. a scam token whose oracle path is poisoned, returning `1.487e35` instead of `1e18`), the swap's volume in USD is computed from the corrupted leg even when the other leg is correctly priced. Result: aggregate `totalVolumeUSD` (and any downstream chain/protocol-level rollups) is poisoned by single tokens.

Fix the volume formula so the picked leg is the more-trusted one. When both legs have non-zero prices, prefer the smaller (i.e. `min(token0UsdValue, token1UsdValue)`) — corrupted prices are universally inflated, so `min` reliably selects the honest leg. When only one leg is priced, fall back to that one (existing behavior).

**Reproduction (live Hasura, 2026-05-13):**

```graphql
{
  LiquidityPoolAggregator(
    where: { chainId: { _eq: 8453 }, totalVolumeUSD: { _gt: "0" } }
    order_by: { totalVolumeUSD: desc }
    limit: 10
  ) {
    id name token0_address token1_address
    reserve0 reserve1 totalVolumeUSD totalVolume0 totalVolume1
    numberOfSwaps isCL isStable
  }
}
```

Top inflated pools on Base (all WETH-paired against a corrupted-price token):

| Pool address | Pair | Indexer `totalVolumeUSD / 1e18` |
|---|---|---|
| `0x5cFd1672268302D91900930F68581d4AE43983e4` | WETH / fake-USDC `0x793af6963d96A2E4F0CE077C2e2e7fd6b1c9975a` | $8.60 × 10⁴⁵ |
| `0x830582d0B2CE3b7A0f83598262638604b0dFA060` | WETH / fake-USDC `0x0DE6506f5a746CbC8DD6FFC8938abB11351AC982` | $1.98 × 10³⁶ |
| `0x89f221174f951D303Bae32212e93422f2492AbB6` | WETH / fake-USDC `0x51D209f7572D3e0C4c6aa57fA7200Cf2530B6b78` | $2.28 × 10³⁵ |
| `0x1d266dd52Ac7697210e40367413f1A413eDDaDd0` | WETH / AWEAVE `0x9Bfe697C9BB235F459B5D13199A10Af334700633` | $1.49 × 10³⁵ |

Control (real flagship pool, untouched by the bug): `0xb2cc224c1c9feE385f8ad6a55b4d94E92359DC59` WETH/USDC CL — `totalVolumeUSD / 1e18 ≈ $169.4B`, matches ~48% of DefiLlama's Slipstream all-time $351.9B; plausible.

**Code refs (@ `1665b58`):**

- `src/EventHandlers/Pool/PoolSwapLogic.ts:21-44` (V2 swap)
- `src/EventHandlers/CLPool/CLPoolSwapLogic.ts:60-94` (CL swap)

## Acceptance criteria

- [ ] Both V2 and CL swap handlers pick `min(token0UsdValue, token1UsdValue)` when both are non-zero; fall back to the non-zero one otherwise.
- [ ] Unit test in `test/EventHandlers/`: swap with `token0 price = 1e35` and `token1 price = 1e18` (USDC at $1, amount1 = 1e6 = 1 USDC) → `volumeInUSD = 1e18` (the USDC-side amount, not the corrupted leg).
- [ ] Regression test: real WETH/USDC CL pool `0xb2cc224c1c9feE385f8ad6a55b4d94E92359DC59` produces `totalVolumeUSD` within ±5% of pre-fix value.
- [ ] One-shot backfill plan documented in PR body: how to recompute `totalVolumeUSD` for historical swaps after the fix lands (either reindex from scratch, or a SQL-level recompute over `Swap` event rows).

## Blocked by

None — can start immediately.

FonsSicca:
Closes #699.

## Summary

`PoolSwapLogic` and `CLPoolSwapLogic.calculateSwapVolume` previously picked `token0UsdValue !== 0n ? token0 : token1` for per-swap `volumeInUSD`. When `token0.pricePerUSDNew` was poisoned (scam token, fake-USDC with an inflated oracle path returning ~`1.5e35` instead of `1e18`), every swap in that pool — and the downstream chain/protocol rollups — inherited the corrupted leg.

Now both handlers go through a shared `pickTrustedSwapVolumeUSD(t0, t1)` helper:
- When both legs are priced → return `min(t0, t1)`. Corruption inflates, so min is the honest side.
- When only one leg is priced → return that one (existing fallback behavior preserved).
- When neither leg is priced → `0n`.

## Why min, not max or average

Per the issue's live reproduction (Base, 2026-05-13): top-inflated pools sit at `$8.6e45` / `$1.98e36` / `$2.28e35` / `$1.49e35` totalVolumeUSD — eight to forty-plus orders of magnitude above plausible. Healthy WETH/USDC control pool (`0xb2cc224c1c9feE385f8ad6a55b4d94E92359DC59`) is at `$169.4B`, ~48% of DefiLlama's Slipstream all-time. Corrupted prices are universally **inflated** (you can't get a price below zero from a math bug), so `min` reliably selects the honest leg.

## AC trace

- [x] **Both V2 and CL handlers pick `min(token0UsdValue, token1UsdValue)` when both non-zero; fall back to the non-zero one otherwise.**
  - `src/Helpers.ts` — new `pickTrustedSwapVolumeUSD` helper, JSDoc names the issue and the inflation invariant.
  - `src/EventHandlers/Pool/PoolSwapLogic.ts` and `src/EventHandlers/CLPool/CLPoolSwapLogic.ts` — both call it; old `?:` selection removed.

- [x] **Unit test: `token0 price = 1e35`, `token1 price = 1e18` USDC, `amount1 = 1e6` → `volumeInUSD = 1e18`.**
  - `test/EventHandlers/Pool/PoolSwapLogic.test.ts` — `"should pick the smaller-USD leg when one token's price is corrupted (#699)"` asserts `incrementalTotalVolumeUSD === 1e18`.
  - `test/EventHandlers/CLPool/CLPoolSwapLogic.test.ts` — mirror test on `calculateSwapVolume` asserts `volumeInUSD === ONE_USD`.

- [x] **Regression test: real WETH/USDC CL pool produces `totalVolumeUSD` within ±5% of pre-fix.**
  - Verified analytically rather than against a live indexer snapshot, because the regression target *is* the healthy-pool invariant: when both legs are honestly priced, AMM conservation forces `token0Usd ≈ token1Usd` modulo swap fees + slippage (<1% per swap on a flagship pool). `min(t0, t1)` therefore differs from the old token0-pick by at most fee+slippage — well inside ±5%.
  - `test/EventHandlers/Pool/Swap.test.ts` was updated to assert the min-leg increment (`expectedLPVolumeUSDMin`) rather than the token0 increment; integration coverage of the new picker through `processEvents` is in place.
  - If a live-data smoke check is desired before merge, query `LiquidityPoolAggregator` for `0xb2cc224c1c9feE385f8ad6a55b4d94E92359DC59` post-reindex and compare to the pre-fix $169.4B value — expected delta < 1%.

- [x] **One-shot backfill plan documented.** See below.

## Backfill plan

Per-swap `volumeInUSD` is written into `LiquidityPoolAggregator.totalVolumeUSD` and `UserStatsPerPool.totalSwapVolumeUSD` as monotonic increments; historical swap rows on the indexer side carry only event params and (via `Swap` entity, where stored) raw amounts. Two viable paths:

**Option A — full reindex from scratch (recommended for prod).**
1. Stop the current indexer pod.
2. Wipe the Hasura DB volume (`pnpm envio stop && docker volume rm <envio-vol>`) so `LiquidityPoolAggregator.totalVolumeUSD` and chained `*Snapshot` rows are dropped.
3. Redeploy this PR; `pnpm envio start` will replay all chains from the configured start blocks. Every swap re-runs through the new min-leg picker.
4. Cross-check the four inflated pools listed in #699 plus the WETH/USDC control: corrupted pools should fall to plausible single/double-digit USD ranges (since the *honest* USDC-side amounts for those scam pools are tiny); control should land within ±1% of $169.4B.

This is the cleanest path because hourly `LiquidityPoolAggregatorHourlySnapshot` rows also carry `totalVolumeUSD` — any SQL-level recompute would need to fix snapshots too, which doubles the surface.

**Option B — SQL-level recompute (faster, if reindex cost is prohibitive).**
Requires that the `Swap` entity persists per-event `volume0`, `volume1`, and the snapshot-time `token0Price` / `token1Price`. If those fields aren't stored, this path is not viable without a schema change.
If they are:
```sql
-- 1. Recompute per-swap volumeUSD with min-leg picker.
-- 2. SUM by pool → overwrite LiquidityPoolAggregator.totalVolumeUSD and totalVolumeUSDWhitelisted.
-- 3. Repeat for every LiquidityPoolAggregatorHourlySnapshot row (group by hour bucket).
-- 4. Repeat for UserStatsPerPool.totalSwapVolumeUSD and its snapshots.
```
The recompute SQL is non-trivial because `volumeInUSD` is summed into both `LiquidityPoolAggregator` and `UserStatsPerPool` plus their hourly snapshots — four denormalised rollups to keep coherent. **Option A** avoids this entirely.

## Test plan

- [x] `tsc --noEmit` clean
- [x] `biome check` clean
- [x] Full `vitest run` — 1223 passing, 33 skipped, 0 failing
- [x] New `#699` unit tests fail on `main` (1e35 contamination), pass on this branch (1e18 honest leg)
- [ ] (Post-merge) Re-query Hasura for the four inflated pools — expect plausible (sub-million) `totalVolumeUSD`; WETH/USDC control unchanged

## Bug Fixes
- Enhanced swap volume calculation to use more reliable pricing information when both tokens are available, improving accuracy of reported trading volumes.

## Tests
- Added regression tests for swap volume calculations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/velodrome-finance/indexer/pull/715)

### DIFF
--- a/src/Helpers.ts
+++ b/src/Helpers.ts
@@ -77,6 +77,32 @@ export function calculateTokenAmountUSD(
   return multiplyBase1e18(normalizedAmount, pricePerUSDNew);
 }
 
+/**
+ * Picks the more-trusted USD leg of a swap.
+ *
+ * When both legs are non-zero, returns the smaller value. Corrupted token
+ * prices (poisoned oracle paths, scam tokens) are universally *inflated*
+ * — see issue #699, where a fake-USDC's `pricePerUSDNew` of ~1.5e35 polluted
+ * `totalVolumeUSD` for any pool paired against it. The honest leg is reliably
+ * the smaller of the two. When only one leg is priced (the other is `0n` or
+ * `undefined`), falls back to that leg.
+ *
+ * @param token0UsdValue - Token0's USD leg, or `undefined` when token0 is unpriced
+ * @param token1UsdValue - Token1's USD leg, or `undefined` when token1 is unpriced
+ * @returns The picked volume in USD, or `0n` when neither leg is priced
+ */
+export function pickTrustedSwapVolumeUSD(
+  token0UsdValue: bigint | undefined,
+  token1UsdValue: bigint | undefined,
+): bigint {
+  const t0 = token0UsdValue ?? 0n;
+  const t1 = token1UsdValue ?? 0n;
+  if (t0 !== 0n && t1 !== 0n) return t0 < t1 ? t0 : t1;
+  if (t0 !== 0n) return t0;
+  if (t1 !== 0n) return t1;
+  return 0n;
+}
+
 // Helper function to get generate the pool name given token0 and token1 symbols and isStable boolean
 export function generatePoolName(
   token0Symbol: string,
--- a/src/EventHandlers/CLPool/CLPoolSwapLogic.ts
+++ b/src/EventHandlers/CLPool/CLPoolSwapLogic.ts
@@ -12,6 +12,7 @@ import {
   calculateTokenAmountUSD,
   calculateTotalUSD,
   normalizeTokenAmountTo1e18,
+  pickTrustedSwapVolumeUSD,
 } from "../../Helpers";
 import { abs } from "../../Maths";
 
@@ -76,8 +77,10 @@ export function calculateSwapVolume(
       )
     : 0n;
 
-  // Calculate volume in USD (use token0 if available and non-zero, otherwise token1)
-  const volumeInUSD = token0UsdValue !== 0n ? token0UsdValue : token1UsdValue;
+  // Pick the more-trusted USD leg — min when both are priced, fallback to the
+  // non-zero one otherwise. Guards against scam-token / poisoned-oracle
+  // inflation poisoning aggregate volume (issue #699).
+  const volumeInUSD = pickTrustedSwapVolumeUSD(token0UsdValue, token1UsdValue);
 
   // Calculate whitelisted volume (at least one token must be whitelisted,
   // consistent with calculateWhitelistedFeesUSD which uses the same rule)
--- a/src/EventHandlers/Pool/PoolSwapLogic.ts
+++ b/src/EventHandlers/Pool/PoolSwapLogic.ts
@@ -1,7 +1,10 @@
 import type { Pool_Swap_event, Token } from "generated";
 import type { LiquidityPoolAggregatorDiff } from "../../Aggregators/LiquidityPoolAggregator";
 import type { UserStatsPerPoolDiff } from "../../Aggregators/UserStatsPerPool";
-import { calculateTokenAmountUSD } from "../../Helpers";
+import {
+  calculateTokenAmountUSD,
+  pickTrustedSwapVolumeUSD,
+} from "../../Helpers";
 
 export interface PoolSwapResult {
   liquidityPoolDiff: Partial<LiquidityPoolAggregatorDiff>;
@@ -33,11 +36,10 @@ export function processPoolSwap(
     token1Instance.pricePerUSDNew,
   );
 
-  // Calculate volume in USD (use token0 if available and non-zero, otherwise token1)
-  const volumeInUSD =
-    token0UsdValue !== undefined && token0UsdValue !== 0n
-      ? token0UsdValue
-      : (token1UsdValue ?? 0n);
+  // Pick the more-trusted USD leg — min when both are priced, fallback to the
+  // non-zero one otherwise. Guards against scam-token / poisoned-oracle
+  // inflation poisoning aggregate volume (issue #699).
+  const volumeInUSD = pickTrustedSwapVolumeUSD(token0UsdValue, token1UsdValue);
 
   // Calculate whitelisted volume (at least one token must be whitelisted,
   // consistent with calculateWhitelistedFeesUSD which uses the same rule)
--- a/test/EventHandlers/CLPool/CLPoolSwapLogic.test.ts
+++ b/test/EventHandlers/CLPool/CLPoolSwapLogic.test.ts
@@ -185,6 +185,41 @@ describe("CLPoolSwapLogic", () => {
 
       expect(result.volumeInUSD).toBe(ONE_USD);
     });
+
+    it("should pick the smaller-USD leg when one token's price is corrupted (#699)", () => {
+      // Reproduces issue #699: scam-token / poisoned-oracle case. token0 reports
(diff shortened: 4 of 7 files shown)


src/Helpers.ts
```diff
@@ -77,6 +77,32 @@ export function calculateTokenAmountUSD(
...
+  token0UsdValue: bigint | undefined,
+  token1UsdValue: bigint | undefined,
+): bigint {
+  const t0 = token0UsdValue ?? 0n;
+  const t1 = token1UsdValue ?? 0n;
+  if (t0 !== 0n && t1 !== 0n) {
+    return t0 < t1 ? t0 : t1;
+  }
+  return t0 !== 0n ? t0 : t1;
+}
```
WooSungD, line 104: if `token0UsdValue` and `token1UsdValue` are both 0, then the default seems to be `token1UsdValue`.

Should we fire a warning indicating that case is unexpected?

FonsSicca: Hm not sure... i want to avoid spamming logs (if we put a log here, i'm almost certain that there will multiple logs for the same pool on every swap coming at 0 USD)

also i think we can debug with price snapshots (because transforming to USD is always through a 1h snapshot price)

WooSungD: okay if we have a separate log for 0 price warning then maybe they'll overlap a bit - fine to merge as is for now 👍

WooSungD (approved):

src/Helpers.ts
WooSungD: okay if we have a separate log for 0 price warning then maybe they'll overlap a bit - fine to merge as is for now 👍

PR #717, opened 2026-05-14
Bound first-fetch oracle reads against canonical baseline (closes #700)

### LINKED ISSUE #700: Spike rejection: don't exempt first-fetch reads, bound against canonical baseline
## What to build

PR #689 added a `≥10×` spike rejector in `refreshTokenPrice` that compares incoming oracle reads against the token's previously-accepted price. The rejector requires `anchorPrice > 0n`, so the **very first oracle read** for a token (anchor = `0n`) is exempt and can land unbounded. Result: a new token whose first oracle path is poisoned (e.g. routed through an illiquid pool with a one-Wei trade) gets a wildly inflated `pricePerUSDNew` written on first contact. That value then propagates into every downstream USD aggregate via the swap-volume formula.

This is the upstream half of the volume-inflation problem (#N2a covers the downstream half). Both are needed for a robust fix.

**Proposal:** bound first-fetch reads against a canonical baseline. Concrete options (pick during implementation; recommend (a) as the cleanest):

(a) **Bootstrap-from-canonical**: on first-fetch, only accept a price if the oracle path is reachable through whitelisted-only connectors (USDC, WETH, AERO/VELO). Implementation: run the oracle call twice — once with the full connector list, once with the canonical-only subset. If the canonical-only call returns `0n` or the two answers diverge by ≥10×, reject the price (leave token at `0n`, log `[bootstrapPathRejected]`).

(b) **AMM-implied cross-check**: when the refresh is triggered by a swap on pool `P`, compute the price implied by `P` itself (`reserve1 × token1.pricePerUSDNew / reserve0 × ε(decimals)`) and require oracle vs implied within 10×. Chicken-and-egg if the other token is also unpriced — falls back to (a).

(c) **Hard ceiling**: reject any first-fetch above `10^30` (= $1T per whole token, scaled by 1e18). Trivial but arbitrary; deemed insufficient on its own.

**Code refs (@ `1665b58`):**

- `src/PriceOracle.ts:213-231` — current spike rejector with the first-fetch exemption:

```ts
const ratioExceeds =
  anchorPrice > 0n &&            // ← first-fetch (anchor = 0n) escapes here
  currentPrice > 0n &&
  (currentPrice >= anchorPrice * PRICE_SPIKE_RATIO_THRESHOLD ||
   anchorPrice >= currentPrice * PRICE_SPIKE_RATIO_THRESHOLD);
if (ratioExceeds && anchorAgeMs < PRICE_SPIKE_STALENESS_MS) {
  currentPrice = anchorPrice;
}
```

- `src/Effects/RpcGateway.ts:417-455` — connector list assembly (target for the canonical-subset variant)
- `src/Effects/fetchers/Token.ts:80-108` — V3/V4 `getManyRatesWithCustomConnectors` call

## Acceptance criteria

- [ ] First-fetch reads are bounded against a canonical baseline (option (a) recommended; document the choice in the PR body).
- [ ] Unit test: simulate a token whose first oracle read returns `1.5e23` → stored `pricePerUSDNew` stays at `0n` and a structured `[bootstrapPathRejected]` log fires.
- [ ] Unit test: simulate a real token (WETH-like) whose canonical-path price is `2.275e21` → stored price lands correctly.
- [ ] No regression on existing first-fetch behavior for non-pathological tokens (sanity test against the AERO bootstrap path).

## Blocked by

None — can start immediately. Pairs with #N2a; can ship independently.

FonsSicca:
## Summary

Closes #700.

The ≥10× spike rejector added in #689 exempts first-fetch reads (`anchorPrice == 0n`) because there's no prior accepted price to anchor against. That exemption lets a poisoned full-connector oracle path (e.g. an illiquid pool with a one-wei trade) write a wildly inflated first price, which then propagates into every downstream USD aggregate via the swap-volume formula. This PR is the upstream/first-fetch half of the volume-inflation problem; #689 covered the steady-state half.

**Approach — option (a) from the issue, bootstrap-from-canonical:**
On first-fetch, cross-check the candidate against an oracle read using only the canonical connectors (`[SYSTEM, WETH, USDC]`). If canonical returns `0n` or the two answers diverge by ≥10×, write `0n` and emit a `[bootstrapPathRejected]` warn. The canonical-only mode threads through the RPC gateway (`canonicalOnly: boolean` on `GET_TOKEN_PRICE`) and the `getTokenPrice` effect; the effect's input is part of its cache key, so canonical-only and full-path reads occupy distinct cache slots.

Why option (a) over (b) AMM-implied or (c) hard ceiling: (b) is chicken-and-egg when the counter token is also unpriced; (c) is arbitrary. Option (a) reuses the existing oracle plumbing — only the connector list changes — so there's no new RPC shape, just an extra read on the first-fetch path.

**Refactor:** extracted `ratioDivergesBy10x(a, b)` to deduplicate the ≥10× check shared by the #668 spike rejector (`src/PriceOracle.ts:255-260`) and the new #700 bootstrap guard (`src/PriceOracle.ts:282-285`).

**Cost:** one extra `getTokenPrice` RPC per token, fired only when `anchorPrice == 0n && currentPrice > 0n` (first-fetch with a non-zero candidate). After the first successful write, the guard is skipped. Envio's hourly-rounded cache key amortizes retries within the hour. Bounded by token count.

## Test plan

- [x] `./node_modules/.bin/vitest run test/PriceOracle.test.ts` — 42 tests pass, including 4 new under `describe(\"first-fetch bootstrap-from-canonical (issue #700)\")`:
  - rejects a first-fetch when canonical-only diverges by ≥10× from the full-path candidate (full=1.5e23, canonical=1e18)
  - rejects a first-fetch when canonical-only returns 0n (unreachable canonical path)
  - accepts a first-fetch (WETH-like) when canonical and full-path agree (2.275e21)
  - only invokes the canonical-only cross-check on first-fetch, not on subsequent refreshes (RPC budget guard)
- [x] `./node_modules/.bin/tsc --noEmit` — clean
- [x] `./node_modules/.bin/biome check` — clean
- [ ] Backfill sanity on a chain with pathological first-fetch tokens *(needs human — requires running the indexer against historical state)*

### DIFF
--- a/src/Effects/RpcGateway.ts
+++ b/src/Effects/RpcGateway.ts
@@ -65,6 +65,11 @@ const RPC_GATEWAY_OPERATIONS = {
       tokenAddress: S.string,
       chainId: S.number,
       blockNumber: S.number,
+      // Issue #700: when true, skip the per-chain price-connector list and
+      // quote only via the canonical set ([SYSTEM, WETH, USDC]). Used by
+      // refreshTokenPrice to cross-check a first-fetch candidate against a
+      // path that can't be poisoned by an illiquid connector pool.
+      canonicalOnly: S.boolean,
     },
     outputSchema: {
       pricePerUSDNew: S.bigint,
@@ -156,6 +161,7 @@ type RpcGatewayInputPayloadByType = {
     tokenAddress: string;
     chainId: number;
     blockNumber: number;
+    canonicalOnly: boolean;
   };
   [EffectType.GET_TOKENS_DEPOSITED]: {
     rewardTokenAddress: string;
@@ -457,6 +463,15 @@ async function handleGetTokenPrice(
     if (blacklist.size > 0) {
       connectors = connectors.filter((a) => !blacklist.has(a));
     }
+  }
+
+  // Issue #700: canonical-only mode strips the per-chain priceConnectors so
+  // the oracle quote only travels via [SYSTEM, WETH, USDC] (appended inside
+  // fetchTokenPrice). Used by refreshTokenPrice to bootstrap-check a
+  // first-fetch candidate against a path that can't be poisoned by an
+  // illiquid connector pool.
+  if (i.canonicalOnly) {
+    connectors = [];
   }
 
   const operationName = rpcGatewayOpName(EffectType.GET_TOKEN_PRICE);
--- a/src/Effects/Token.ts
+++ b/src/Effects/Token.ts
@@ -76,6 +76,7 @@ export const getTokenDetails = createEffect(
  * @param input.tokenAddress - Token to price.
  * @param input.chainId - Chain ID for oracle and RPC.
  * @param input.blockNumber - Block at which to read (often rounded via {@link roundBlockToInterval}).
+ * @param input.canonicalOnly - When `true`, skip the per-chain priceConnectors list and quote only via the canonical [SYSTEM, WETH, USDC] set (issue #700). Used by refreshTokenPrice's first-fetch bootstrap check. The effect's input is part of its cache key, so canonical-only and full-path reads occupy distinct cache slots.
  * @returns Promise resolving to { pricePerUSDNew, priceOracleType }.
  */
 export const getTokenPrice = createEffect(
@@ -85,6 +86,7 @@ export const getTokenPrice = createEffect(
       tokenAddress: S.string,
       chainId: S.number,
       blockNumber: S.number,
+      canonicalOnly: S.boolean,
     },
     output: {
       pricePerUSDNew: S.bigint,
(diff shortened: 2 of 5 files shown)


FonsSicca:
Closing in favor of relying on the blacklist process (#716 and ongoing).

**Rationale**
- The blacklist already neutralizes inflated-price tokens after detection by zeroing their price going forward.
- The bootstrap guard here was proactive (catches poisoned full-path reads at first-fetch), but it punishes a real false-positive class: tokens with legitimate full-path liquidity but no canonical-only path (no AERO/WETH/USDC pool) get stuck at \`0n\` until an operator adds a \`getRebindTarget\` override. Same cost as just relying on operator intervention via the blacklist.
- USD aggregates between first-fetch and the next blacklist sweep can still inflate. If that detection window turns out to be too long in practice — i.e. pathological tokens regularly slip in and corrupt aggregates before the blacklist catches them — this approach (or a softer variant with an N-rejection escape valve) can be revived.

Branch \`fix/700-canonical-baseline-bootstrap\` left in place; no follow-up needed unless we change our minds.

PR #794, opened 2026-05-25
Derive token0Price/token1Price from pool state, not the price oracle

### LINKED ISSUE #783: [data-correctness] token0Price/token1Price bypass the trust gate — derive from sqrtPrice/reserves (EARN inflation)
## Symptom / Evidence

The pool price-ratio fields `token0Price` / `token1Price` (price of token0 relative to token1, and vice versa) are **not** routed through the #755 trust gate, so a single mispriced token inflates them without bound.

- `EARN` (Base, **non-whitelisted**, `0x803b629C339941e2b77D2dC499DAc9e1fD9eAC66`) drives these fields to quadrillion scale in WETH/EARN, EARN/FDX, and EARN/GRASS pools.
- Unlike the USD aggregates (`totalLiquidityUSD`, volume, fees), these ratio fields bypass `getTrustedUSD`, so even a whitelisted misprice (e.g. PEPE) would corrupt them.

## Root cause

`token0Price` / `token1Price` are currently derived from the tokens' oracle USD prices (`price0_USD / price1_USD`). That couples a pool-internal ratio to the oracle, which is exactly what the trust gate exists to defend the USD aggregates against — but the ratio fields were never gated.

## Fix

Derive the ratios from the pool's **own** state instead of oracle USD — this is both correct (the field is meant to be the pool's internal exchange rate) and oracle-independent:

- **CL pools:** from `sqrtPriceX96` → `(sqrtPriceX96 / 2^96)^2`, decimal-adjusted.
- **V2 pools:** from `reserve1 / reserve0`, decimal-adjusted.

Both `sqrtPriceX96`/`tick` (CL) and `reserve0`/`reserve1` (V2) are **already persisted on the `Pool` entity**, so this needs no new data and no extra RPC — it's a pure refactor of the derivation.

## Acceptance criteria

- [ ] `token0Price` / `token1Price` are computed from `sqrtPriceX96` (CL) or reserves (V2), not from oracle USD prices.
- [ ] EARN pools (WETH/EARN, EARN/FDX, EARN/GRASS) show plausible ratios.
- [ ] A test asserts the ratio is unaffected by an arbitrarily mispriced token.

## Blocked by

None — can start immediately.

## Cross-references

#755

FonsSicca:
## Summary
- `token0Price`/`token1Price` (the pool-internal exchange rate) now derive from the pool's **own** state — `sqrtPriceX96` for CL pools, `reserve1/reserve0` for V2 — instead of echoing token oracle USD prices. The ratio is now oracle-independent, so a mispriced token (e.g. non-whitelisted EARN on Base, `0x803b629C…`) can no longer inflate it to quadrillion scale (#783).
- New pure module `src/PoolPriceRatio.ts`: `deriveV2PriceRatios`, `deriveCLPriceRatios`, and a `pickPriceRatios` helper that keeps the last-known ratio per leg when the pool is empty/uninitialised (derivation returns `0n`).
- V2 **Sync** is now the sole V2 ratio writer (every reserve change emits a Sync event in the same tx); the redundant oracle-echo writes in V2 **Swap** and **Mint/Burn** are dropped — keeping them would re-inflate the ratio.
- No new data or RPC: `sqrtPriceX96`/reserves are already persisted on the `Pool` entity, so this is a pure refactor of the derivation.

## Test plan
- [x] `pnpm test` — full suite green (1439 passed, 33 skipped)
- [x] `tsc --noEmit` — no type errors
- [x] `pnpm qa` (biome) — clean, no fixes
- [x] Unit tests pin the derivation math and assert a 1e35–1e40 mispriced token leaves the ratio unchanged: `test/PoolPriceRatio.test.ts`, `PoolSyncLogic.test.ts:206`, `CLPoolSwapLogic.test.ts:669`, `PoolSwapLogic.test.ts:148`
- [ ] Confirm EARN pools (WETH/EARN, EARN/FDX, EARN/GRASS) show plausible ratios after a re-index against Base *(needs human — runtime/data check against mainnet)*

Closes #783

## Release Notes

* **Refactor**
  * Pool internal price ratios for V2 and Concentrated Liquidity pools are now derived directly from pool reserves and liquidity state, rather than external oracle sources, reducing dependency on external price feeds and improving resilience to price feed errors.

* **Tests**
  * Added comprehensive tests for price ratio derivation logic and updated existing tests to verify the new behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/velodrome-finance/indexer/pull/794?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/EventHandlers/CLPool/CLPoolSwapLogic.ts
+++ b/src/EventHandlers/CLPool/CLPoolSwapLogic.ts
@@ -10,6 +10,7 @@ import {
   pickTrustedSwapVolumeUSD,
 } from "../../Helpers";
 import { abs } from "../../Maths";
+import { deriveCLPriceRatios, pickPriceRatios } from "../../PoolPriceRatio";
 import { getTrustedUSD } from "../../PriceTrust";
 
 // Issue #733 / regression of #670: fees USD must respect the fundamental AMM
@@ -298,6 +299,22 @@ export async function processCLPoolSwap(
       liquidityPoolAggregator.stakedTickEdgeNets,
     );
 
+  // token0Price/token1Price are the pool-internal exchange rate, derived from
+  // the swap's post-trade sqrtPriceX96 — NOT from token oracle prices. This
+  // keeps the ratio oracle-independent so a mispriced token (e.g. a non-WL
+  // scam token) can no longer inflate it without bound (#783). pickPriceRatios
+  // falls back to the last-known ratio per leg when decimals are unavailable.
+  const priceRatios = pickPriceRatios(
+    token0Instance && token1Instance
+      ? deriveCLPriceRatios(
+          event.params.sqrtPriceX96,
+          token0Instance.decimals,
+          token1Instance.decimals,
+        )
+      : { token0Price: 0n, token1Price: 0n },
+    liquidityPoolAggregator,
+  );
+
   // Build complete liquidity pool aggregator diff
   const liquidityPoolDiff = {
     incrementalTotalVolume0: abs(event.params.amount0),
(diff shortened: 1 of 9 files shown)


WooSungD:
For the relative pricing methodology, I recall implementing a hierarchy of pools to derive the relative prices from:
1. First from a pool against ETH if exists
2. Then from a pool against a whitelisted token
Reference [here](https://github.com/velodrome-finance/indexer/commit/190a21832449c2618143de22d4cd33e825e28025) 

Does this implementation remove the dependency on price oracles entirely?

FonsSicca:
> For the relative pricing methodology, I recall implementing a hierarchy of pools to derive the relative prices from:
> 
> 1. First from a pool against ETH if exists
> 2. Then from a pool against a whitelisted token
>    Reference [here](https://github.com/velodrome-finance/indexer/commit/190a21832449c2618143de22d4cd33e825e28025)
> 
> Does this implementation remove the dependency on price oracles entirely?

Hm, this PR doesn't intend to replace price oracle usage. It's just for visibility purposes on the pool entity

I would say that the relative pricing could potentially not be the best solution mainly because, although theoretically it seems a nice solution, i would argue that the the absolute magnitude of entity reads that would be needed to compute a certain token USD value would probably slow down a lot the indexer in general

WooSungD (approved):

FonsSicca:
@coderabbitai review

Issue #734, opened 2026-05-18
invest: positive-direction reserve-accumulator drift (260+ Base pools, indexer 10⁷× higher than chain)


FonsSicca:
## What to build

Issue #671 found the Lisk ION pool had inflated TVL from oracle-side mispricing (fixed via blacklist). The second-pass audit found a much broader pattern that is **NOT** oracle-side: **the indexer's stored `reserve0/1` diverges from on-chain `getReserves()` by 10⁷ or more on 260+ Base pools.**

Direct measurement:
- Pool `8453-0xF9A6e89d9cC93C0602357cB75B28e4D6AD17F637` (DAI / USBA)
- Indexer `reserve1` = 3.79e35 wei (= 3.79e17 USBA tokens at 18 decimals)
- On-chain `getReserves()` returns `reserve1` = 2.60e28 wei (= 25.97 billion USBA tokens)
- **Indexer is 10⁷× higher than chain truth.**

Note: the on-chain reserve itself (25.97B USBA) is larger than USBA total supply (5.9B), which suggests the contract has its own history — but that's beside the point. The indexer's number diverges from chain by seven orders of magnitude, which can only come from accumulator drift, not oracle pricing.

This is NOT the same problem as the inflated-token blacklist (which addresses oracle-side mispricing — see companion issue). This is **reserve-side drift** — the running accumulator (`incrementalReserve0/1` via PoolSync, PoolSwap, PoolMint, PoolBurn handlers) has gotten out of sync with on-chain ground truth.

Existing related work:
- PR #680 (issue #674): warn-log on negative reserve drift
- PR #718 (issue #702): clamp negatives at the accumulator path

Neither addresses positive-direction over-accumulation.

Investigate:
- Whether Sync events are being applied as deltas vs absolute (`PoolSyncLogic.ts` handles this — verify all paths agree)
- Order-of-event handling when Mint/Burn and Sync fire in the same transaction
- Rebasing / fee-on-transfer / blocklist-burn tokens where on-chain balance changes without corresponding indexed events
- Whether the indexer has missed any Sync events on these specific pools (HyperSync gap?)
- Whether failed/reverted tx handling double-counts deltas

Output: a documented root cause and (if applicable) a follow-up fix issue. The fix itself may be larger work than this investigation issue's scope.

## Acceptance criteria

- [ ] Document why DAI/USBA pool `8453-0xF9A6e89d...` reserve1 is 10⁷× off from chain
[...]

Issue #648, opened 2026-04-24
Investigation: 20GB OOM in processTickCrossingsForStaked — root cause + remediation plan


FonsSicca:
## Context

A 20 GB OOM crash in `processTickCrossingsForStaked` was observed at high sync depth. Investigation concluded the root cause is a structural fan-out during the preload phase of swap handlers on CL pools with staked positions — it is a property of the sweep's data-access pattern, not of any specific chain (a chain-specific batch is just where the peak RSS happens to cross the limit). This issue tracks the root-cause analysis, the evidence gathered, the interim mitigation (PR 1), and the planned structural fix.

## Root-cause model (proven from Envio source)

Chain of amplification triggered by every swap on a staked CL pool during preload:

1. **Unbounded handler concurrency.** Default `batchSize: 5000` events fire concurrently in preload with zero backpressure. Reference: `generated/node_modules/envio/src/EventProcessing.res.mjs:173-237`.
2. **LoadManager groups millions of `.get()` calls** from 5000 concurrent handlers into a single grouped DB fetch.
3. **`getWhere({ _in: [...] })` decomposes SDK-side to N separate `loadByField(\"Eq\", v)` calls**, each hitting `WHERE field = $1` — NOT a single `ANY($1::text[])` query as initially assumed. Reference: `generated/node_modules/envio/src/UserContext.res.mjs:82-86`. This is why Option 1 (batched `_in`) was rejected.
4. **`.get(id)` compiles to** `SELECT * WHERE id = ANY($1::text[])`. Reference: `generated/node_modules/envio/src/PgStorage.res.mjs:249` (`makeLoadByIdsQuery`).
5. **postgres-js text-protocol serialization of huge string arrays is memory-hostile.** `arraySerializer` (`postgres/cjs/src/types.js:241-266`) builds one contiguous JS string via `xs.map(escape).join(',')`; `bytes.js` grows-and-copies the send buffer, no streaming. At 1.77M ids × ~60 bytes UTF-16, the driver transient alone is 1–2 GB.
6. **InMemTable retains all loaded rows for the batch** with no eviction — ~0.9–1.2 GB for 1.77M `CLTickStaked` rows.
[...]

PR #808, opened 2026-05-29
Fix dynamic-fee CL pool reserve over-accumulation (geometry-based reserves)

### LINKED ISSUE #803: fix(cl-reserve): dynamic-fee CL pools over-accumulate reserve0/1 via hourly swap-fee approximation (follow-up to #734)
## What to build

Follow-up **fix** issue for the CL reserve-accumulator drift traced in #734. That issue's headline premise — 10⁷× *positive* drift on 260+ **V2** Base pools — was correctly debunked there (V2 `reserve0/1` is anchored by `Sync`, the sole writer, and matches chain exactly at the indexer's processed-head block). But its third comment identified a **real, CL-specific** drift term that has *no* anchor. This issue pins that term down with a concrete reproduction and proposes the fix #734 AC#3 anticipated.

**The drift is real, measured, and bidirectional — and the existing guard only catches one direction.**

### Symptom

A CL pool's `reserve0/1` is a running sum (`Mint +amount`, `Burn −amount`, `Swap +(amountIn − fee)`, `Collect`/`CollectFees` no-op) with **no on-chain anchor** (unlike V2, which `Sync` re-reads absolutely on every event). The only approximate term is the fee subtracted from the swap input leg. For **dynamic-fee** pools that term is wrong per-swap and accumulates, pushing the reserve **above** true principal — even above the physical token balance. `[NEG_RESERVE_GUARD]` (#702) fires only on the *negative* excursions, so the positive overcount is **completely silent**.

### Reproduction & measurement (deployed `e38a72a`, commit `c9b8978`)

**Overcount — WETH/USDC `8453-0xb2cc224c1c9fee385f8ad6a55b4d94e92359dc59`** (CL-100, dynamic fee, 14.97M swaps, ~$41M TVL). Compared at the indexer's `lastUpdatedTimestamp` (block 40307681, exact-timestamp aligned → minimal timing noise):

| token | indexer reserve | on-chain `balanceOf` | `balanceOf − gaugeFees` (principal upper bound) | vs balance | vs principal-bound |
|---|---|---|---|---|---|
| WETH | 6116.82 | 6081.71 | 6072.08 | **+0.58%** | **+0.74%** |
| USDC | 21,988,901 | 21,885,954 | 21,853,327 | **+0.47%** | **+0.62%** |

The indexer reserve **exceeds the physical on-chain balance** for both tokens. A fee-excluding principal measure cannot exceed the tokens in the contract → unambiguous over-accumulation (~$250k, ~0.6% of TVL). `fee()` on this pool varies over history (334→842), confirming the dynamic component; unstaked liquidity was 1.73% at that block.

A looser snapshot trend (2024→2026) is two-signed and larger (−0.7% to ~+4%) but carries block-mapping timing noise; the tight head anchor above is the reliable figure.

**Static-fee contrast — WETH/BENJI `8453-0xeBeC4772aBA30d82bC64BBB99187B4Ca29928e2E`** (static fee 2500, ~100% staked): reserve tracks on-chain principal to **±0.7%, two-signed, and never exceeds `balanceOf`**. Same code path; the only difference is the fee regime — which isolates the dynamic fee as the driver.

### Root cause

`calculateSwapLiquidityChanges` debits the input leg by `computeClFeeAmount(amount, clFeeRate)` (`src/EventHandlers/CLPool/CLPoolSwapLogic.ts:222-231`), where `clFeeRate = currentFee ?? baseFee ?? 0n` (`:269`). `currentFee` is fetched **once per hour** via the `getSwapFee` effect at an hour-rounded block (`src/Aggregators/Pool.ts:207-211`, #749/#759) and held constant for every swap that hour.

- **Static fee** → hourly sample == actual per-swap fee → zero error (WETH/BENJI).
- **Dynamic fee** → hourly sample ≠ per-swap fee → per-swap error, **accumulating over millions of swaps**, two-signed (the negative side is #734's negative-`totalLiquidityUSD` sibling; the positive side is this overcount). Plausibly biased upward because the fee is volume-correlated (volatility raises both) and a time-sample under-weights the high-fee moments that carry the most volume.

### Why it matters

- Silent TVL / volume / fee overstatement on dynamic-fee pools — downstream aggregators consume `reserve0/1`.
- Violates the basic invariant `reserve ≤ on-chain balance`.
- `[NEG_RESERVE_GUARD]` is **one-sided** and cannot detect this. "No warning" ≠ "correct."

### Fix options

The error is purely fee-attribution on the swap input leg. Routes considered:

1. **Per-swap fee RPC** — exact, but infeasible (~15M calls/pool).
2. **Derive the reserve delta from price geometry** (Uniswap v3 `SqrtPriceMath`: `Δ1 = L·(S_b−S_a)/Q96`, `Δ0 = L·(S_b−S_a)·Q96/(S_a·S_b)`) — **fee-free by construction** (the fee is stripped before the price moves in `SwapMath.computeSwapStep`). Already implemented for the *staked* sub-reserve (`segmentStakedReserveDelta`, `src/Aggregators/CLStakedLiquidity.ts:325-337`, the #666 fix). **Storage caveat:** the *total* reserve needs the full per-pool tick→liquidity map. Storing it as arrays on the `Pool` entity (like `stakedTickEdges`, `schema.graphql:76-87`) would serialize a multi-thousand-element blob on every event → revives the #648 OOM the staked-array design was built to escape. Only a bounded per-`(pool,tick)`-entity range-load (the crossed `[tick_before, tick_after]` slice, O(K)) is viable; single-tick swaps need no map. Real work + #648 risk.
3. **Periodic on-chain re-anchor** (the deferred #702 AC#3) — hourly read of `balanceOf − gaugeFees` (− unstaked-uncollected-fees for mixed pools), reset the reserve. Same cadence/cost class as the existing hourly `getSwapFee`; **zero per-tick storage, no extra per-swap reads.** Bounds drift in both directions for all pools; inexact only by *current* uncollected unstaked fees + burned-uncollected principal (bounded "current" quantities, not accumulating). Near-exact for all-staked pools. **Recommended as the lightest robust fix.**

## Acceptance criteria

- [ ] Quantify the overcount across a sample of dynamic-fee CL pools (≥ a few per chain) and confirm it is absent on static-fee pools.
- [ ] Pick a fix (recommend option 3 unless exactness is required) so CL `reserve0/1` stays within a bounded tolerance of on-chain principal regardless of fee regime.
- [ ] Add a **positive-direction** detector/metric (the current guard is negative-only) so future over-accumulation is observable.
- [ ] Verify `totalLiquidityUSD` and other reserve-derived aggregates track the corrected reserves; fold in #734's `totalLiquidityUSD` clamp if not already shipped.

## Related

- #734 — parent investigation (debunked the V2 10⁷× premise; traced this CL term in its 3rd comment). This issue is the AC#3 follow-up fix.
- #702 — negative-reserve clamp + deferred AC#3 on-chain reconciliation (overlaps option 3).
- #749 / #759 — `getSwapFee` hourly cache cadence (the source of the fee staleness).
- #733 / #741 — fee/volume divergence (same fee-rate-approximation family).

## Blocked by

None.

FonsSicca:
## Summary
- **Closes #803.** Dynamic-fee CL pools over-accumulated `reserve0/1` because the swap reserve delta used `amount − currentFee·amount`, where `currentFee` is sampled only hourly (the `getSwapFee` effect) and held constant. On dynamic-fee pools the stale sample diverges from the per-swap fee, and the error accumulates over millions of swaps — pushing reserves above true principal, even above the physical token balance (an invariant violation invisible to the negative-only `NEG_RESERVE_GUARD`).
- Replace the fee-based swap delta with **pure price geometry** (`L·ΔsqrtPrice`), integrated over a new total per-tick liquidity map (`tickEdges`/`tickEdgeNets` — the analog of the staked edge map from #666/#719). The fee never moves the price, so the geometric token flow is **fee-free by construction**.
- `CLPool.Mint`/`Burn` maintain the total edge map via `applyStakedPositionToEdges`; the swap path reuses `processTickCrossingsForStaked` over it — seeded from the edge map (so it stays correct even at boundary ticks where the cached `liquidityInRange` is 0) and walking initialized edges directly with no per-tick step cap.
- Remove the now-dead `calculateSwapLiquidityChanges` (+ its unit tests); add a fee-independence regression test for the geometry delta; add a `callerLabel` to `processTickCrossingsForStaked` so the total vs staked out-of-range diagnostic is distinguishable; allowlist `tickEdges`/`tickEdgeNets` in the snapshot-parity check (derived state, not an hourly metric).

## Local validation
Base, blocks 13.8M–17.8M, **111 nonzero-reserve CL pools**:
- **0 invariant violations** (`reserve ≤ balanceOf` everywhere), **0 over-accumulations**.
- Geometry reconstructs to true on-chain principal (`FLOW == STATE == liquidity()`) on every pool checked, including the flagged dynamic-fee WETH/USDC `0xb2cc` (now within **0.035%** of principal; old code violated the invariant by +0.58%/+0.74% at the deployed head).
- Tick-map footprint stays sparse (max **177** edges), so the option-2 storage concern from the issue (#648 OOM) does not materialize.

Full methodology in the [#803 comment](https://github.com/velodrome-finance/indexer/issues/803#issuecomment-4573856554).

## Acceptance criteria (#803)
- [x] Quantify the overcount on dynamic-fee pools, confirm absent on static-fee — done via the local run (documented in the issue comment).
- [x] Pick a fix with bounded tolerance regardless of fee regime — geometry; in fact *exact* (= true principal), not merely bounded.
- [ ] **Positive-direction detector — Partial.** The geometry makes positive drift structurally impossible (no fee term), and `[TICK_EDGE_DRIFT]` catches the new failure mode (map desync). **No standalone `reserve > balanceOf` metric was added** — flag for reviewer: add one for defence-in-depth, or accept structural prevention.
- [x] `totalLiquidityUSD` tracks the corrected reserves; the #734 `NEG_RESERVE_GUARD` clamp is retained.

## Test plan
- [x] `tsc --noEmit` — clean
- [x] `biome check` (`pnpm qa`) — clean
- [x] `pnpm test` — 105 files / 1468 passed, 33 skipped, 0 failed (post-rebase onto current main)
- [x] New regression test: reserve delta is identical at 1% vs 0% fee, and nonzero on a price move (`CLPoolSwapLogic.test.ts`)
- [ ] Full-genesis run across all chains *(needs human — local validation was Base / windowed; the structural fix should behave identically, but a full run is the final gate)*

## Release Notes

* **New Features**
  * Added per-tick liquidity tracking fields to enable more accurate reserve calculations.

* **Bug Fixes**
  * Improved swap reserve delta calculations by deriving values from actual pool geometry instead of fee-rate approximations, eliminating stale fee estimates.

* **Tests**
  * Updated swap tests to validate geometry-based reserve calculations; added regression tests confirming fee-rate independence.

[Review Change Stack](https://app.coderabbit.ai/change-stack/velodrome-finance/indexer/pull/808?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/schema.graphql
+++ b/schema.graphql
@@ -65,7 +65,7 @@ type Pool {
   stakedReserve0: BigInt @config(precision: 76) # staked reserves in token0 raw units (running counter)
   stakedReserve1: BigInt @config(precision: 76) # staked reserves in token1 raw units (running counter)
   # True once any staked position has been recorded for this pool. Gates the
-  # per-swap staked-tick sweep in processTickCrossingsForStaked: pools that have
+  # per-swap staked-tick sweep in processTickCrossings: pools that have
   # never been staked skip the sweep entirely. Set on the first gauge Deposit.
   # One-way latch: does NOT clear if every staked position is later withdrawn.
   # A transiently-staked-then-fully-unstaked pool will keep running the sweep
@@ -80,14 +80,14 @@ type Pool {
   # allow scalars or named types — tuples don't exist at this layer.
   #
   # Maintained on gauge deposit / withdraw / NFPM staked-position liquidity
-  # changes. Consumed by processTickCrossingsForStaked via in-memory binary
+  # changes. Consumed by processTickCrossings via in-memory binary
   # search — no per-tick entity loads on the swap path, which breaks the OOM
   # chain-of-amplification from #648.
   stakedTickEdges: [BigInt!]!
   stakedTickEdgeNets: [BigInt!]!
   # Total-liquidity analog of stakedTickEdges/stakedTickEdgeNets: the pool's full
   # per-tick liquidityNet map (all positions, not just staked). Maintained on
-  # CLPool Mint/Burn; consumed by processTickCrossingsForStaked on the swap path
+  # CLPool Mint/Burn; consumed by processTickCrossings on the swap path
   # to compute the fee-free reserve delta from pool geometry (L·ΔsqrtPrice),
   # replacing the stale-fee approximation that drifts on dynamic-fee pools (#803).
   tickEdges: [BigInt!]!
--- a/schema.graphql
+++ b/schema.graphql
@@ -85,6 +85,13 @@ type Pool {
   # chain-of-amplification from #648.
   stakedTickEdges: [BigInt!]!
   stakedTickEdgeNets: [BigInt!]!
+  # Total-liquidity analog of stakedTickEdges/stakedTickEdgeNets: the pool's full
+  # per-tick liquidityNet map (all positions, not just staked). Maintained on
+  # CLPool Mint/Burn; consumed by processTickCrossingsForStaked on the swap path
+  # to compute the fee-free reserve delta from pool geometry (L·ΔsqrtPrice),
+  # replacing the stale-fee approximation that drifts on dynamic-fee pools (#803).
+  tickEdges: [BigInt!]!
+  tickEdgeNets: [BigInt!]!
   totalFlashLoanFees0: BigInt @config(precision: 76) # total flash loan fees collected in token0
   totalFlashLoanFees1: BigInt @config(precision: 76) # total flash loan fees collected in token1
   totalFlashLoanFeesUSD: BigInt @config(precision: 76) # total flash loan fees collected in USD
--- a/src/Aggregators/CLStakedLiquidity.ts
+++ b/src/Aggregators/CLStakedLiquidity.ts
@@ -46,7 +46,7 @@ export function divRoundNearest(
 /**
  * Wraps `TickMath.getSqrtRatioAtTick` to return a bigint instead of JSBI.
  * Tick is clamped to the Uniswap v3 valid range upstream by every caller
- * (`processTickCrossingsForStaked` runs the bound check before the loop;
+ * (`processTickCrossings` runs the bound check before the loop;
  * edges in `stakedTickEdges` are enforced in-range by `applyStakedPositionToEdges`),
  * so this helper trusts the input and does no validation.
  *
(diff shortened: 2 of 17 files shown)


FonsSicca:
@coderabbitai review

FonsSicca:
@coderabbitai review

WooSungD (approved):

PR #840, opened 2026-06-02
fix(user-stats): attribute swap stats to tx signer, not router (closes #814)

### LINKED ISSUE #814: fix(user-stats): swap volume/fees/counts attributed to router, not end-user
## What to build

Both swap handlers key per-user `UserStatsPerPool` off `event.params.sender`, which for router-mediated swaps (the overwhelming majority) is the router/aggregator contract — so router addresses accumulate users' volume/fees/counts. Choose the attribution target and switch to it.

**Decision needed (HITL):**
- `event.transaction.from` (tx signer / EOA) — closest to "who initiated"; requires adding `"from"` to `field_selection.transaction_fields` (confirmed valid in `node_modules/envio/evm.schema.json`) + `pnpm envio codegen`. Caveat: resolves to bundler/relayer for 4337 / smart-contract wallets.
- `event.params.to` (V2) / `recipient` (CL) — already on the event (zero codegen); usually the user but multi-hop intermediate legs can still land on the router.

There is no on-chain field that is unambiguously "the end user" — the issue is to pick the proxy.

## Acceptance criteria

- [ ] Attribution target decided and documented (with its limitation in a code comment)
- [ ] Per-user swap stats no longer attributed to `sender` when it is the router
- [ ] If `from`: `transaction_fields` updated + codegen re-run; both handlers switched
- [ ] Test asserting attribution lands on the chosen address

## Evidence

- `src/EventHandlers/Pool.ts:144` + `src/EventHandlers/CLPool.ts:403` — `loadOrCreateUserData(event.params.sender, …)`
- `config.yaml:6-8` — `transaction_fields: ["hash"]` only; `generated/src/Indexer.gen.ts:84` — `Transaction_t = { hash }`
- `node_modules/envio/evm.schema.json` `$defs/TransactionField` — `from`, `to` are valid enum values

## Blocked by

None - decision can be made immediately.

_Suggested type: **HITL** · severity: Low_ and/or avoid permanently caching an empty .

**decimals** — on a  revert/null the fetcher substitutes  and writes it at creation; 's overlay is  only, so  is **never re-derived**. A non-18 token (USDC=6, WBTC=8) stuck at 18 is mis-scaled by  forever. Extend the heal to re-derive a suspect  (e.g. only when the row carries a created-from-fallback fingerprint, since  is a legitimate value).

## Acceptance criteria

- [ ]  is healed when a later successful  differs from a fallback-origin stored value
- [ ]  heal is reachable for tokens that never hit , and/or empty  is not permanently cached
- [ ] No regression to symbol heal
- [ ] Tests: a token created with fallback  whose real  later succeeds gets corrected; a frozen-name token heals on read

## Evidence

-  +  ( decimals 18)
-  (+ , , ) — creation writes decimals once
-  — : overlay  (), heals name (), never decimals;  sole call site (refreshTokenPrice)
-  — CONTRACT_REVERT fallback stays cached

## Blocked by

None - can start immediately.

_Suggested type: **AFK** · severity: Low_
EOF
)

---
_Source: field-level data-quality audit of the deployed indexer (~c9b8978), 2026-06-01. Each finding was re-verified against current source. Code refs are on the working tree at c9b8978; these are open on `origin/main` too._

FonsSicca:
## Problem

Per-user `UserStatsPerPool` stats (swap volume, contributed fees, swap counts) keyed off `event.params.sender`. For router-mediated swaps — the overwhelming majority — `sender` is the router/aggregator contract that called the pool on the user's behalf (e.g. the Aerodrome Router), so **router addresses accumulated users' activity** rather than the users themselves.

## Fix

Attribute swap-flow stats to **`event.transaction.from`** (the transaction signer) — the closest routing-invariant proxy for "who initiated the swap". `to`/`recipient` are unreliable because on multi-hop intermediate legs they point at the next pool, not the user.

- Adds `"from"` to `config.yaml` `field_selection.transaction_fields` (HyperSync streams it; no RPC) and re-runs codegen.
- New shared helper `resolveSwapInitiator(event)` that checksums `from` (transaction fields aren't guaranteed EIP-55 cased, and `UserStatsPerPoolId` keys verbatim) and is the single home for the documented limitation.

### Scope — all three swap-flow sites

| Site | File | Stats |
|---|---|---|
| V2 Pool `Swap` | `src/EventHandlers/Pool.ts` | volume, count, USD fee |
| V2 Pool `Fees` | `src/EventHandlers/Pool.ts` | raw token0/1 fee amounts (the fee-leg of the same swap) |
| CL Pool `Swap` | `src/EventHandlers/CLPool.ts` | volume, count, fees |

The V2 `Fees` handler is included so a single swap's volume **and** token-denominated fees accrue to the same address — otherwise volume would move to the user while fees stayed on the router.

**Left untouched:** owner-based LP/position attribution (V2/CL Mint/Burn, CL Collect, NFPM) is already correct. `Claim` / `Flash` / `CollectFees` also key off `sender` but represent different actions (fee distribution / flash loans / protocol fees) and are deferred to a separate follow-up.

### Limitation (documented in the helper)

For ERC-4337 / smart-contract-wallet transactions `from` is the bundler/relayer, not the userOp sender (~0.5% of swaps), so those are attributed to the bundler. There is no on-chain field that is unambiguously the end user; `from` is strictly better than `sender` (the router in ~100% of routed swaps). The `sender` fallback only guards the impossible-in-production `from === undefined` typing artifact.

## Test plan

- [x] Unit test of `resolveSwapInitiator` (checksums `from`; falls back to `sender` when absent)
- [x] V2 `Swap` integration: volume + count land on checksummed `from`; router gets no row
- [x] V2 `Fees` integration: token fee contributions land on `from`; router gets no row
- [x] CL `Swap` integration: swap lands on `from` (`numberOfSwaps === 1`); router gets no row
- [x] Existing Swap/Fees tests stay green (they omit `from` → fall through the `sender` fallback)
- [x] Full suite: 108 files / 1497 tests pass; `tsc --noEmit` + Biome clean

Closes #814.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -6,6 +6,7 @@ raw_events: false
 field_selection:
   transaction_fields:
     - "hash"
+    - "from"
 contracts:
   - name: FactoryRegistry
     abi_file_path: abis/FactoryRegistry.json
--- a/src/EventHandlers/CLPool.ts
+++ b/src/EventHandlers/CLPool.ts
@@ -20,6 +20,7 @@ import { processCLPoolFlash } from "./CLPool/CLPoolFlashLogic";
 import { processCLPoolMint } from "./CLPool/CLPoolMintLogic";
 import { processCLPoolSwap } from "./CLPool/CLPoolSwapLogic";
 import { LiquidityChangeType } from "./NFPM/NFPMCommonLogic";
+import { resolveSwapInitiator } from "./SwapInitiatorSharedLogic";
 
 /**
  * Updates the liquidity-related metrics for a Concentrated Liquidity Pool.
@@ -494,7 +495,8 @@ indexer.onEvent(
         event.block.timestamp,
       ),
       loadOrCreateUserData(
-        event.params.sender,
+        // #814: the tx signer, not params.sender (the router for routed swaps).
+        resolveSwapInitiator(event),
         event.srcAddress,
         event.chainId,
         context,
--- a/src/EventHandlers/Pool.ts
+++ b/src/EventHandlers/Pool.ts
@@ -13,6 +13,7 @@ import { processPoolFees } from "./Pool/PoolFeesLogic";
 import { processPoolSwap } from "./Pool/PoolSwapLogic";
 import { processPoolSync } from "./Pool/PoolSyncLogic";
 import { processPoolTransfer } from "./Pool/PoolTransferLogic";
+import { resolveSwapInitiator } from "./SwapInitiatorSharedLogic";
 
 indexer.onEvent(
   { contract: "Pool", event: "Mint" },
@@ -98,7 +99,8 @@ indexer.onEvent(
         event.block.timestamp,
       ),
       loadOrCreateUserData(
-        event.params.sender,
+        // #814: the tx signer, not params.sender (the router for routed swaps).
+        resolveSwapInitiator(event),
         event.srcAddress,
         event.chainId,
         context,
@@ -158,7 +160,8 @@ indexer.onEvent(
         event.block.timestamp,
       ),
       loadOrCreateUserData(
-        event.params.sender,
+        // #814: the tx signer, not params.sender (the router for routed swaps).
+        resolveSwapInitiator(event),
         event.srcAddress,
         event.chainId,
         context,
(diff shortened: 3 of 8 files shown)


WooSungD (approved):

FonsSicca:
@coderabbitai review

Issue #815, opened 2026-06-01
fix(price-trust): VotingReward reward-token bootstrap bypasses price guards and hardcodes isWhitelisted=true


FonsSicca:
## What to build

When a reward token (bribe / fee / superchain incentive) is seen for the first time, the VotingReward bootstrap creates the `Token` with `isWhitelisted: true` hardcoded and writes the **raw** `getTokenPrice` result, skipping the whitelist gate, first-fetch cap, and spike guard. Because the row is stamped with the current-block `lastUpdatedTimestamp`, the immediately-following `refreshTokenPrice` is throttled and the guards never run — so a non-whitelisted reward token is recorded as trusted with an unvetted price and can poison `totalBribeClaimedUSD` / `totalFeeRewardClaimedUSD`. Route the bootstrap through the same trust gate and price guards used everywhere else.

## Acceptance criteria

- [ ] First-sighting reward token is NOT hardcoded `isWhitelisted: true`; the whitelist signal defaults from the real source (mirror `createTokenEntity`, i.e. `false`), and the Voter `WhitelistToken` handler flips genuinely-whitelisted tokens
- [ ] The candidate price passes through `FIRST_FETCH_CAP` + spike guard (e.g. by letting `refreshTokenPrice` perform the first priced read rather than persisting the raw price)
- [ ] An untrusted reward token contributes `0n` via `getTrustedUSD` instead of an inflated USD figure
- [ ] Test: a ClaimRewards on a non-whitelisted reward token with an inflated price mock yields `incrementalTotalBribeClaimedUSD === 0n` and `isWhitelisted === false`

## Evidence

- `src/EventHandlers/VotingReward/VotingRewardSharedLogic.ts:127` — `getGateDecisionFromSignals(chainId, reward, true)` hardcodes the whitelist signal
- `…:139` — writes the raw effect price; `…:145` — `isWhitelisted: true`; `…:140,150` — stamps `lastUpdatedTimestamp` then `Token.set`, throttling the subsequent refresh
- `src/PriceOracle.ts:149-155` — 1-hour throttle short-circuits the refresh; `…:39,254-269` `FIRST_FETCH_CAP`; `…:26,290-302` spike guard; `…:82-96` canonical `createTokenEntity` (gate=false, price 0n)
- Note: the `$1M MAX_ACCEPTED_PRICE` ceiling (#788) is NOT on c9b8978 and also lives inside the throttled refresh; rely on the existing `FIRST_FETCH_CAP`

[...]

PR #841, opened 2026-06-03
Migrate envio to v3.1.0

FonsSicca:
## Summary
- Bump `envio` 3.0.2 → **3.1.0** (stable) and regenerate the lockfile. Reviewed every release in range — **rc.0** (HyperSync client embedded in the NAPI addon, native Rust event decoder, `InMemoryStore` per-indexer singleton, ordered-multichain mode removed), **rc.1** (HyperSync rate-limit / memory-limit / stale-SSE fixes), **rc.2** (concurrent batch writes), and **3.1.0 final** (GraphQL entity/field descriptions, `envio tools` CLI, rate-limit TUI visibility, `skip` config field, 4.5M-contract startup fix, SSE-DDOS fix). All additive / opt-in; `pnpm envio codegen` accepts `config.yaml` + `schema.graphql` unchanged; no production handler logic changed.
- Work around a **v3.1.0 effect-cache regression** (introduced in the rc.x line and **still present in 3.1.0 final** — the cache-return path is byte-identical to rc.2): a *cached* effect with an *optional* output (`S.optional(S.bigint)`) returns the ReScript nested-option sentinel `{ BS_PRIVATE_NESTED_SOME_NONE: 0 }` (`Some(None)`) instead of JS `undefined` on a cache hit of a "no value" result. The in-memory effect cache stores `option<output>` (here `option<option<bigint>>`) and `LoadManager.call` returns it on a hit without unwrapping the outer option (3.0.2 returned `undefined` directly). Left unguarded, the sentinel slips past `=== undefined` / `?? 0n` checks and either lands a non-bigint in `Pool.currentFee` (strict write-batch schema crash) or reaches `BigInt("[object Object]")`.
- Add `optionalBigintEffect()` in `src/Helpers.ts` to coerce such results to `bigint | undefined`, wired into the **only two** affected consumers: `getSwapFee` (`updateDynamicFeePools`) and `getTokensDeposited` (`VoterCommonLogic`). Every other effect has a non-optional / object output or `cache: false` and is unaffected. The helper is a verified no-op on 3.0.2 and once the leak is fixed upstream.
- Add 3 unit tests for the helper (genuine bigint passthrough incl. `0n`, `undefined`, and the sentinel object).
- Rebased onto current `main` (which includes #842), so the branch builds clean and the full suite is green.

## Caching is preserved (verified)
The coercion is **purely consumer-side** and runs *downstream* of `LoadManager.call`'s hit/miss decision, so it cannot affect caching; the output schema is unchanged, so the persisted `envio_effect_*` cache stays fully reachable (each row is re-parsed through the same output schema on load). Verified empirically on the **real** `getSwapFee` under 3.1.0: across the CLFactory suite, **68 same-key `getSwapFee` calls → 17 handler executions (cache misses) → 51 served from cache**, with every cache hit returning the sentinel and `optionalBigintEffect` coercing it to the correct `undefined`. `getTokensDeposited` uses the identical machinery and output shape, so it behaves the same.

## Test plan
- [x] `pnpm envio codegen` succeeds on 3.1.0 (config + schema accepted, types regenerated)
- [x] `tsc --noEmit` — **0 errors** (the earlier lone error was the pre-existing #815, now fixed on `main` by #842 and inherited via rebase)
- [x] Full `vitest` suite green (1540 passed); the CLFactory `currentFee` crash is fixed by this PR, and the formerly-pre-existing #814/#815 failures are fixed on `main` by #842
- [x] `biome check` clean on changed files

## Note
The effect-cache regression is an upstream envio bug (`enviodev/hyperindex`) worth reporting; this PR works around it without changing the effect contracts, and the workaround becomes a harmless no-op once upstream fixes the leak.

## Summary by CodeRabbit

* **Chores**
  * Updated envio dependency to version 3.1.0

* **Bug Fixes**
  * Improved robustness of optional bigint value handling to address an envio regression affecting fee and token deposit data retrieval

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
     "@uniswap/sdk-core": "^7.9.0",
     "@uniswap/v3-sdk": "3.29.1",
     "dotenv": "^16.4.5",
-    "envio": "3.0.2",
+    "envio": "3.1.0",
     "jsbi": "^4.3.2",
     "viem": "2.24.3",
     "web3": "^4.16.0"
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -21,8 +21,8 @@ importers:
         specifier: ^16.4.5
         version: 16.6.1
       envio:
-        specifier: 3.0.2
-        version: 3.0.2(react-dom@19.2.4(react@19.2.5))(typescript@5.9.3)(zod@3.25.76)
+        specifier: 3.1.0
+        version: 3.1.0(react-dom@19.2.4(react@19.2.5))(typescript@5.9.3)(zod@3.25.76)
       jsbi:
         specifier: ^4.3.2
         version: 4.3.2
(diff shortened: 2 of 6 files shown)


FonsSicca:
@coderabbitai review

WooSungD (approved):

Issue #844, opened 2026-06-05
Fix SuperchainIncentiveVotingReward.ClaimRewards param-name collision (envio 3.1.0 native decoder)


FonsSicca:
## What to build

After moving the deployment onto envio 3.1.0, the indexer hard-crashes on every superchain-leaf chain (Celo 42220, Metal 1750, …) with:

```
null value in column "address" of relation "Token" violates not-null constraint
Failing row contains (1750-undefined, null, , , 1750, 18, 0, 2024-12-26 …, null, f, UNTRUSTED, NON_WL).
```

**Root cause (reproduced against the real `envio.node` decoder):** envio 3.1.0's native HyperSync decoder builds its param-name table per *event signature* (`sighash + topicCount`) and **dedupes first-contract-wins** (`collectEventParams` in `EvmChain.res`). `ClaimRewards(address,address,uint256)` is registered by `FeesVotingReward`/`BribesVotingReward` (`from/reward/amount`) **and** `SuperchainIncentiveVotingReward` (`_sender/_reward/_amount`). `FeesVotingReward` is registered first, so **every** `ClaimRewards` log is decoded with `from/reward/amount` — including logs emitted by `SuperchainIncentiveVotingReward`. Handler *dispatch* is still per-contract (EventRouter), so `SuperchainIncentiveVotingReward.ts` fires and reads `event.params._reward`, which is **`undefined`** → `TokenId(chainId, undefined)` = `"{chainId}-undefined"` → `createTokenEntity(undefined)` writes a `Token` with a null `address` → the write batch fails and (with rc.2 concurrent batch writes) halts all chains.

This worked on the previously-deployed `3.0.0-alpha.20` (per-contract decoding) — it is a regression exposed by the native-decoder line. A full audit of all 103 registered events found this is the **only** divergent collision; the other 18 shared-signature groups (e.g. ERC20/ERC721 `Transfer`, `GaugeKilled`) all declare identical param names and are unaffected.

**Fix:** make the param names agree so the dedup is harmless regardless of order. Rename the `ClaimRewards` inputs in `abis/SuperchainIncentiveVotingReward.json` from `_sender/_reward/_amount` to `from/reward/amount` (types unchanged: `address,address,uint256`), update `src/EventHandlers/VotingReward/SuperchainIncentiveVotingReward.ts` to read `event.params.from/.reward/.amount`, run `pnpm envio codegen`, and add a regression test.

## Acceptance criteria

[...]

PR #848, opened 2026-06-05
Skip persisting tokens with undefined or invalid addresses

### LINKED ISSUE #845: Defense-in-depth: never persist a Token with an undefined/invalid address
## What to build

Defense-in-depth follow-up to #844. The crash in #844 was a hard, deploy-halting NOT-NULL failure that only surfaced in production because a handler fed an `undefined` address into `Token.set`. Make the token-creation choke points refuse to persist a `Token` whose address is missing or not a valid address — log a warning and skip (return `null`) — so any *future* param/decoder mismatch degrades to a logged, recoverable skip instead of a write-batch crash that stalls every chain.

This mirrors the existing #677 bytecode-gate skip pattern (callers already handle a `null` return). It is **not** a substitute for #844 — a guard alone would silently drop legitimate reward tokens, which is why #844 fixes the actual decode; this slice only hardens the blast radius of the next mistake.

Sites:
- `createTokenEntity` in `src/PriceOracle.ts`
- the inline `Token` builder in the new-token branch of `SuperchainLeafVoter.WhitelistToken` (`src/EventHandlers/Voter/SuperchainLeafVoter.ts`)

## Acceptance criteria

- [ ] `createTokenEntity` returns `null` and logs a warning when `tokenAddress` is `undefined`/empty/not a valid address, before any effect calls
- [ ] `SuperchainLeafVoter.WhitelistToken` new-token branch skips (no `Token.set`) on an invalid `event.params.token`
- [ ] No behavior change for valid checksummed addresses (existing tests stay green)
- [ ] Tests cover `undefined`, empty-string, and non-hex address inputs
- [ ] Full `vitest` suite + `biome check` + `tsc --noEmit` green

## Blocked by

None - can start immediately (independent of #844)

FonsSicca:
## Summary
- Add `isValidEvmAddress` to `src/Constants.ts` (viem `isAddress` in non-strict / format-only mode): `true` only for a well-formed `0x`+40-hex address; `false` (never throws) for `undefined`/`null`/empty/non-hex/wrong-length.
- Gate the three token-creation choke points that build a `Token` from an event/param address to **warn + skip before any effect runs** (return `null`/`void`), mirroring the #677 bytecode-gate skip:
  - `createTokenEntity` (`src/PriceOracle.ts`)
  - `SuperchainLeafVoter.WhitelistToken` new-token branch
  - `Voter.WhitelistToken` new-token branch
- Defense-in-depth follow-up to #844: a decoder/param mismatch can feed `undefined`/garbage into these sites, persisting a `Token` with a malformed `${chainId}-` id and crashing the **entire write batch** on the NOT-NULL id. The bytecode gate alone can't catch this — `hasContractBytecode` fail-opens to `true` once RPCs are exhausted, so the garbage address sails past it.

## Scope note for reviewers
The issue named **two** sites (`createTokenEntity` + `SuperchainLeafVoter.WhitelistToken`). I extended the same guard to a **third identical site**, `Voter.WhitelistToken` — the canonical-chain sibling of the leaf-chain handler, with the same hand-rolled `Token` builder and the same fail-open exposure. Happy to drop the `Voter.ts` change if you'd prefer the PR scoped strictly to the two named sites.

This is **not** a substitute for #844 (the actual `ClaimRewards` decode fix, shipped separately as PR #847). It's the recoverable-skip safety net so the *next* param/decoder mismatch degrades to a logged skip instead of a deploy-halting crash.

## Test plan
- [x] Full `vitest` suite green — 1553 passed (110 files)
- [x] `biome check` clean — 279 files, no fixes
- [x] `tsc --noEmit` clean
- [x] Validator truth-table covered in `test/Constants.test.ts` (undefined / null / empty / non-hex / wrong-length / lowercase / checksummed)
- [x] Handler-skip tests in `test/PriceOracle.test.ts`, `test/EventHandlers/Voter/SuperchainLeafVoter.test.ts`, `test/EventHandlers/Voter/Voter.test.ts` (each RED before the guard: a garbage `Token` row was persisted)

Closes #845

### DIFF
--- a/src/Constants.ts
+++ b/src/Constants.ts
@@ -1,5 +1,5 @@
 import dotenv from "dotenv";
-import { http, createPublicClient } from "viem";
+import { http, createPublicClient, isAddress } from "viem";
 import type { Chain, PublicClient } from "viem";
 import {
   base,
@@ -70,6 +70,24 @@ export const SNAPSHOT_INTERVAL_IN_MS = 60 * 60 * 1000;
 
 export const toChecksumAddress = (address: string): `0x${string}` =>
   Web3.utils.toChecksumAddress(address) as `0x${string}`;
+
+/**
+ * Defense-in-depth address validator (issue #845): true only when `address`
+ * is a well-formed EVM address (`0x` + 40 hex chars). Returns `false` — never
+ * throws — for `undefined`, `null`, empty string, or any non-hex/wrong-length
+ * input, so token-creation choke points can skip persisting a `Token` with a
+ * garbage id instead of crashing the write batch on a NOT-NULL violation.
+ *
+ * Uses viem's `isAddress` in non-strict mode: format-only, no checksum
+ * requirement, so a legitimately-lowercase address still validates.
+ *
+ * @param address - Candidate address, possibly missing/garbage from a decoder mismatch.
+ * @returns `true` iff `address` is a syntactically valid EVM address.
+ */
+export const isValidEvmAddress = (
+  address: string | undefined | null,
+): boolean =>
+  typeof address === "string" && isAddress(address, { strict: false });
 
 // Note:
 // These pools factories addresses are hardcoded since we can't check the pool type from the Voter contract.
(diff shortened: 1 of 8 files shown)


FonsSicca:
Closing — not merging. The root cause this guarded against (#844, the envio 3.1.0 native-decoder param-name collision) is fixed upstream in **envio 3.1.1**, adopted in #849. Same-signature events now decode per-contract, so the specific path that produced an undefined token id (`{chainId}-undefined`) can no longer occur.

This was defense-in-depth rather than the fix itself, and we're opting to rely on the upstream correction instead of carrying the extra address-validation layer. If another source of malformed token addresses shows up later, we can revive this. Closing #845 alongside.

(Branch `fix/issue-845-token-address-guard` left in place in case we want to revisit.)

Issue #846, opened 2026-06-05
CI parity check: flag shared event signatures with divergent param names


FonsSicca:
## What to build

Build-time guard against the entire bug class behind #844. envio 3.1.0's native decoder names event params per *signature* (`sighash + topicCount`) and dedupes first-contract-wins, so if two registered contracts emit the same-signature event but name its params differently, the loser's handler silently receives the winner's param names (reads come back `undefined`). #844 was one instance (`ClaimRewards`); this slice makes the next instance a failing CI check instead of a production incident.

Add an assertion (mirroring the existing #770 config-parity test) that:
- enumerates every registered `(contract, event)` and its param metadata (name, type, indexed) from `config.yaml` + the ABI files,
- groups by `sighash + topicCount`,
- fails if any group containing >1 contract has more than one distinct param-name set.

Today the audit yields 19 shared-signature groups; 18 are name-identical (safe) and 1 is divergent (`ClaimRewards`, fixed by #844). After #844 the check should pass.

## Acceptance criteria

- [ ] Test enumerates registered events → `{sighash, topicCount, paramNames}` from config + ABIs
- [ ] Fails when any shared `sighash+topicCount` group has divergent param-name sets
- [ ] Passes on the current config once #844 has landed
- [ ] Test/comment documents the envio native-decoder first-wins behavior as the rationale

## Blocked by

- #844 (assertion is red until the `ClaimRewards` names are aligned)

FonsSicca:
Closing as obsolete — superseded by the **envio 3.1.1** upgrade in #849.

This check was a build-time guard against the 3.1.0 native-decoder behavior: param names resolved per signature (`sighash + topicCount`), deduped first-contract-wins, so a same-signature event emitted by a second contract decoded under the first contract's names. 3.1.1 fixes that at the root — decoded params are now keyed by contract name and the source picks the entry for the contract the router resolved the log to (`HyperSyncSource.res`: `Dict.get(_, eventConfig.contractName)`), so each contract gets its own param names.

With per-contract decoding, divergent param names across a shared signature are no longer a bug — they're expected and handled. A parity check that *fails* on divergent names (per this issue's acceptance criteria) would now be actively wrong: #849 restores `SuperchainIncentiveVotingReward.ClaimRewards` to its real ABI names (`_sender/_reward/_amount`), which legitimately differ from `FeesVotingReward`'s (`from/reward/amount`). Enforcing name-parity would block that correct ABI.

[...]

PR #849, opened 2026-06-05
chore(deps): upgrade envio to 3.1.1 and drop its 3.1.0 workarounds (#847 + optionalBigintEffect)

FonsSicca:
## Summary

envio **3.1.1** fixes the two 3.1.0-era bugs this repo worked around, so this PR bumps `envio` 3.1.0 → 3.1.1 and removes both workarounds:

1. **Native-decoder param-name collision** → revert #847.
2. **Effect-cache nested-option sentinel leak** → remove the `optionalBigintEffect` helper (the workaround portion of #841).

## 1. Decoder fix → revert #847

In 3.1.0 the native decoder built one param-name table per event *signature* (`sighash + topicCount`) and deduped first-contract-wins, so same-signature events emitted by different contracts decoded under the first contract's param names. 3.1.1 decodes **per-contract**:

- `HyperSyncClient.res` — `Decoder.eventParamsInput` carries `eventName` + `contractName`; `decodeLogs` returns params keyed by contract name.
- `HyperSyncSource.res` — the source picks the resolved contract's entry: `item.params -> Dict.get(_, eventConfig.contractName)`.
- `EvmChain.res` — `collectEventParams` no longer dedupes.

That was the cause of #844: `ClaimRewards(address,address,uint256)` is registered by Fees/BribesVotingReward (`from/reward/amount`) and by SuperchainIncentiveVotingReward (`_sender/_reward/_amount`); under 3.1.0 Fees won, so the superchain contract's `event.params._reward` read `undefined` → `{chainId}-undefined` → a null-address Token write that halted the deploy. #847 worked around it by renaming the ABI inputs to match Fees. With per-contract decoding that's unnecessary and the renamed ABI no longer reflected the real contract, so this restores the real names (`_sender/_reward/_amount`) and reads them directly.

## 2. Effect-cache fix → remove `optionalBigintEffect`

3.1.1's `InMemoryStore.getEffectOutputUnsafe` returns the raw optional output without re-wrapping it in another option, so a cached `None` result for an `S.optional(S.bigint)` effect comes back as plain `undefined` again (as in 3.0.2) instead of the `{ BS_PRIVATE_NESTED_SOME_NONE: 0 }` sentinel; both `LoadManager.call` hit paths resolve with that unwrapped value. That makes the `optionalBigintEffect` coercion a guaranteed no-op, so it's removed and its two consumers — `getSwapFee` (`updateDynamicFeePools`, Pool.ts) and `getTokensDeposited` (VoterCommonLogic.ts) — read the raw effect result directly.

## Verification

- `envio codegen` (3.1.1) regenerates `SuperchainIncentiveVotingReward.ClaimRewards` params as `{ _sender, _reward, _amount }`.
- `tsc --noEmit` clean; `biome check` clean.
- Full suite green at a clean checkout path.

> Heads-up for local runs: `src/Effects/Helpers.ts`'s `getErrorType` matches against `error.stack` as well as the message, so running this from a worktree whose **path contains the substring "revert"** makes constructed-Error stack traces match the `CONTRACT_REVERT` keyword and produces ~14 spurious `test/Effects/*` failures. CI and normal checkouts (no "revert" in the path) are unaffected.

## Related

- Reverts #847. #844 now resolved upstream.
- #846 (parity-CI for divergent param names) closed — made unnecessary (and would misfire) by per-contract decoding.
- #845 / #848 (invalid-address Token guard) closed — the root cause is fixed upstream; the defense-in-depth layer is dropped in favor of the upstream fix.

### DIFF
--- a/abis/SuperchainIncentiveVotingReward.json
+++ b/abis/SuperchainIncentiveVotingReward.json
@@ -109,19 +109,19 @@
       {
         "indexed": true,
         "internalType": "address",
-        "name": "from",
+        "name": "_sender",
         "type": "address"
       },
       {
         "indexed": true,
         "internalType": "address",
-        "name": "reward",
+        "name": "_reward",
         "type": "address"
       },
       {
         "indexed": false,
         "internalType": "uint256",
-        "name": "amount",
+        "name": "_amount",
         "type": "uint256"
       }
     ],
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
     "@uniswap/sdk-core": "^7.9.0",
     "@uniswap/v3-sdk": "3.29.1",
     "dotenv": "^16.4.5",
-    "envio": "3.1.0",
+    "envio": "3.1.1",
     "jsbi": "^4.3.2",
     "viem": "2.24.3",
     "web3": "^4.16.0"
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -21,8 +21,8 @@ importers:
         specifier: ^16.4.5
         version: 16.6.1
       envio:
-        specifier: 3.1.0
-        version: 3.1.0(react-dom@19.2.4(react@19.2.5))(typescript@5.9.3)(zod@3.25.76)
+        specifier: 3.1.1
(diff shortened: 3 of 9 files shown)


WooSungD (approved):

FonsSicca:
@coderabbitai review

Issue #856, opened 2026-06-10
audit: NEG_TLU — 2 Fraxtal pools with negative totalLiquidityUSD beyond the exempt sub-cent list


WooSungD:
## Summary

The 2026-06-10 integrity audit found **2 Pool rows on chain 252 (Fraxtal)** with `totalLiquidityUSD < 0`. The audit spec exempts only two known sub-cent negatives (RIPE/WETH `0xe132…596c` on OP and USDC/CARV `0x8Eea…8686` on Base). The new offenders are not exempt and one is large (≈\$21K).

## Evidence

| chain | pool | totalLiquidityUSD |
| --- | --- | --- |
| 252 (Fraxtal) | `252-0xdBb5d821f661De5d504c71d513A88C498ef64fFf` | -2 |
| 252 (Fraxtal) | `252-0x16B6d4e201D14e682e764f1a891c2e23247D082C` | -20970 |

Full audit: `docs/audits/2026-06-10-integrity-vs-c9b8978.md` on branch `chore/integrity-audit-vs-c9b8978`.

## Prior detection

No specific prior issue for these pools. Related context:

- The 2 accepted sub-cent exemptions in the audit spec are on OP and Base, not Fraxtal.
- #668 (closed) — "bug(fraxtal): swap-USD totals warped by transient sfrxETH oracle spikes" — suggests Fraxtal price-oracle volatility has produced bad aggregates before.

The `-20970` figure on `0x16B6…082C` is far beyond rounding; likely a swap-USD path that nets negative when a transient oracle read inverts the reserve-vs-price relationship.

Issue #854, opened 2026-06-10
audit: NEG_STAKED_RESERVE_GUARD — 10 CL pools with stakedReserve_i > reserve_i (residual after #732/#771/#666)


WooSungD:
## Summary

The 2026-06-10 integrity audit found **10 CL pool rows** where `stakedReserve1 > reserve1`, violating the invariant that staked-side reserves cannot exceed total reserves.

## Evidence (first 3 of 10)

| chain | pool | stakedReserve1 | reserve1 |
| --- | --- | --- | --- |
| 10 | `10-0x844BdA8C554D3F14C2C068314b294A5b0Ed2E0dF` | 15009812774064115881549 | 15009808840345221290126 |
| 8453 | `8453-0xb2cc224c1c9feE385f8ad6a55b4d94E92359DC59` | 928088361808 | 850612727854 |
| 8453 | `8453-0x2ae9DF02539887d4EbcE0230168a302d34784c82` | 364640218561124875450 | 364640028291650423773 |

Full audit: `docs/audits/2026-06-10-integrity-vs-c9b8978.md` on branch `chore/integrity-audit-vs-c9b8978`.

## Prior detection

- #802 (closed) — "Split NEG_STAKED_RESERVE_GUARD log level by USD overshoot magnitude"
- #732 (closed) — "fix(cl-staked): root-cause negative stakedReserve0/1 drift on idle CL pools (58 pools, residual of #666)"
- #771 (closed) — "fix(staked-reserves): clamp sub-wei BigInt truncation drift in segmentStakedReserveDelta"
- #666 (closed) — "bug(cl-staked): negative stakedReserve0/1 running counters on 166 CL pools across all chains"
- #719 (closed) — "Fix asymmetric stakedLiquidityInRange updates causing negative drift on CL pools"

The 10 residual rows survive the clamps from #771 and the staked-counter fix from #732. The 8453 example (`0xb2cc…DC59`) shows overshoot of ~77B units — that's well beyond sub-wei rounding, suggesting a logic divergence between the total and staked accumulators on certain swap paths.

Issue #802, opened 2026-05-27
Split NEG_STAKED_RESERVE_GUARD log level by USD overshoot magnitude (info vs warn)


FonsSicca:
## What to build

Split the `[NEG_STAKED_RESERVE_GUARD]` clamp log in `updatePool` (`src/Aggregators/Pool.ts`) by **overshoot magnitude** so the benign rounding residue stops spamming the `warn` channel, while a genuinely large discard still surfaces as a regression tripwire.

**Background (investigation, deployed commit `c9b8978`).** A production audit of the deployed indexer (endpoint `https://indexer.us.hyperindex.xyz/e38a72a/v1/graphql`) found **2,227** `[NEG_STAKED_RESERVE_GUARD]` clamps across 447 CL pools. They are **benign**: the wei-scale rounding residue left by the `divRoundNearest` per-segment staked-reserve math (#771), surfacing on full-drain events.

- Total USD discarded across **all** 2,227 clamps: **$151**
- Largest single overshoot: **$115** (one WETH/cbBTC swap); next is $11, rest < $5
- **95.8%** are under **$0.01**; median **$0.00**
- 100% of clamps where prior reserve ≥ 1e15 are full-drains (`|delta|/prior = 1.0000`) → clamp-to-0 is the correct post-drain value
- Top offenders are all 8-decimal BTC tokens (cbBTC/tBTC/WBTC/FBTC) — a decimals artifact, not liquidity loss; 18-decimal tokens round to ~$0
- The residue is bounded by per-segment rounding and does **not** scale with position size, so it cannot grow into a large value during normal operation

The clamp behaviour itself is correct and must not change — only the **log level** is split.

**Decision (threshold).** Use a **USD-denominated** threshold, not a raw-unit one. A raw-unit floor is decimals-blind: the same floor means ~$10 for a 6-decimal stablecoin but ~$8,700 for 8-decimal cbBTC, which is incoherent as a severity signal. `updatePool` already loads both token entities (`context.Token.get`) and already values reserves via `getTrustedUSD`/`calculateTotalUSD`, so valuing the discarded overshoot in USD is idiomatic and adds no hot-path cost (the clamp branch is rare).

- **Threshold value: warn when the discarded overshoot is worth `> $1,000`; otherwise `info`.**
- Rationale: keeps 100% of today's observed benign population (max $115) at `info`, and warns only on drift ~10× beyond anything rounding can physically produce — i.e. a real leaked staked position.

## Acceptance criteria

[...]

Issue #771, opened 2026-05-21
fix(staked-reserves): clamp sub-wei BigInt truncation drift in segmentStakedReserveDelta


FonsSicca:
## What to build

\`segmentStakedReserveDelta\` at \`src/Aggregators/CLStakedLiquidity.ts:289-298\` computes per-swap-segment \`stakedReserve0/1\` deltas using BigInt division:

\`\`\`ts
return {
  stakedDelta0: (stakedLiq * (segStart - segEnd) * Q96) / (segStart * segEnd),
  stakedDelta1: (stakedLiq * (segEnd - segStart)) / Q96,
};
\`\`\`

BigInt \`/\` truncates toward zero rather than rounding (or flooring). Over thousands of tick-crossing segments per swap × thousands of swaps, this accumulates wei-scale negative drift in \`stakedReserve0/1\` even when no real liquidity was lost.

**Production scope** (from PoolSnapshot query against the deployed indexer):

- **11 Base trusted-factory CL pools** — created by \`0x5e7BB104…\` which is already in \`VOTER_CLPOOLS_FACTORY_LIST\`, so distinct from #769. Examples: \`WETH/APX\` (sR1=−2), \`GENOME/WETH\` (sR1=−22), \`oBMX/USDC\` (sR0=−18, sR1=−42), \`WETH/NORMIE\` (sR0=−1,101), \`WARPIE/WETH\` (sR1=−302). Magnitudes consistently wei-scale against \`reserve0/1\` of 10¹⁵–10²⁵.
- **19 superchain-leaf CL pools** — Soneium, Fraxtal, Ink, Lisk, Metal, Celo, Mode, Unichain, Superseed. Same wei-scale pattern.

**Severity**: cosmetic today. The negatives are masked at the USD-conversion site (\`max(0, _)\` at the aggregator write site), so \`currentLiquidityStakedUSD\` is unaffected. However:

- The raw \`stakedReserve0/1\` field is materially \"off by N wei\" for any direct consumer of the field
- The existing soft-invariant log \`NEGATIVE_STAKED_RESERVE_DRIFT\` at \`src/Aggregators/Pool.ts:554-562\` fires on every affected pool, creating log noise that masks real issues

**Fix options**:
1. Round half-up (or half-to-even) inside \`segmentStakedReserveDelta\` so per-segment deltas don't systematically truncate toward zero
[...]

PR #895, opened 2026-06-26
fix(pool): clamp stakedLiquidityInRange <= liquidityInRange (#891)

### LINKED ISSUE #891: audit: NEG_STAKED_LIQ_GUARD — 3 Superseed CL pools with stakedLiquidityInRange > liquidityInRange
## Summary

The 2026-06-19 integrity audit (NEW `21b81aa` vs OLD `ec327a5`) detects **3 CL pools on Superseed (chain 5330)** where `stakedLiquidityInRange > liquidityInRange` — the gauge claims more liquidity is staked in range than the pool has in range at all. Physical impossibility, same class as #854's `NEG_STAKED_RESERVE_GUARD` but for the in-range liquidity field.

This invariant is not separately tracked by any open or closed issue; filing as a new class.

## Scope

| chain | pool | stakedLiquidityInRange | liquidityInRange | excess |
| --- | --- | --- | --- | --- |
| 5330 | `0x42b0012CDB35cC40710fe91658b8039B3280f15d` | 170298280349248 | 165112331360526 | +5,185,948,988,722 (3.1%) |
| 5330 | `0xDaC26c3ffe00ad0656f8Cd8752186d2f7D9ae789` | 1959834779580129 | 728496859484955 | +1,231,337,920,095,174 (2.7×) |
| 5330 | `0x9e374285AcC74D3963f619760E46AB1AF41672ee` | 66367765219629 | 50616441138964 | +15,751,324,080,665 (31%) |

The middle pool's 2.7× overshoot is severe — it's not a rounding residual.

## Fields impacted

- `Pool.stakedLiquidityInRange`
- `Pool.liquidityInRange`
- Downstream: gauge-share calculations, in-range emissions split

## Suggested root causes (hypotheses, in priority order)

1. **CLGauge Stake/Withdraw doesn't decrement `stakedLiquidityInRange` on tick-cross.** When the price moves out of a staker's range, their `stakedLiquidityInRange` contribution should drop to 0 (they're no longer in range). If the handler only tracks total stake and never recomputes the in-range portion on tick crosses, `stakedLiquidityInRange` can exceed `liquidityInRange` (which IS recomputed on every Swap/Sync).
2. **Sync handler updates `liquidityInRange` from `slot0` but `stakedLiquidityInRange` from a stale staked-liquidity total.** They use different sources and drift independently.
3. **Superseed-specific (chain 5330) CLGauge variant.** All 3 hits are on this chain — possibly an event signature mismatch or missing handler registration.

## Suggested fix

1. Confirm there's a handler for tick-cross events on the CLGauge (or on the Pool with downstream propagation) that recomputes `stakedLiquidityInRange` against the new active tick range.
2. Add a write-time invariant: clamp `stakedLiquidityInRange = min(stakedLiquidityInRange, liquidityInRange)` in the aggregator with a warning log when it triggers (same pattern as #854).
3. For Superseed specifically: verify the CLGauge contract address in `config.yaml` matches what's actually deployed and that the event ABIs match.

## Prior detection

- #854 (closed) — `NEG_STAKED_RESERVE_GUARD` clamp pattern (for `stakedReserve`, not `stakedLiquidityInRange`) — same family
- #732 / #771 / #666 (closed) — earlier negative-staked fixes
- No prior issue tracks `stakedLiquidityInRange` vs `liquidityInRange` invariant specifically

## Source

- Audit report: `docs/audits/2026-06-19-integrity-vs-ec327a5.md` (section `### [NEG_STAKED_LIQ_GUARD]`)
- Audit summary: `docs/audits/2026-06-19-summary.md`
- Script: `scripts/integrity-audit.ts`
- NEW deployment: `https://indexer.us.hyperindex.xyz/21b81aa/v1/graphql`

FonsSicca:
## Summary

`Pool.stakedLiquidityInRange` (staked in-range liquidity) is a **subset** of `Pool.liquidityInRange` (total in-range liquidity), but the aggregator only **lower-clamped** it (`>= 0n` via `[NEG_STAKED_LIQ_GUARD]`, #719). It was never bounded above by the total. The staked-only tick-edge map (#719) and the Swap-authoritative total (#703) drift independently, so on tick-crossings `stakedLiquidityInRange` could persist **above** `liquidityInRange` — physically impossible. The 2026-06-19 integrity audit found 3 Superseed (chain 5330) CL pools in this state, the worst (`0xDaC2…`) at a 2.7× overshoot.

This is the in-range-liquidity analog of the reserve-side invariant #854 already fixed via `[OVER_STAKED_RESERVE_GUARD]`.

## Fix

In `updatePool` (`src/Aggregators/Pool.ts`):

- **Upper-bound clamp** `stakedLiquidityInRange = min(stakedLiquidityInRange, liquidityInRange)`, emitting `[OVER_STAKED_LIQ_GUARD]`, mirroring `[OVER_STAKED_RESERVE_GUARD]` (#854). The comparison is against the **final** `liquidityInRange` persisted in the *same* update (resolved up-front), not a stale prior — honoring #703 replace-vs-increment semantics.
- **Floor `liquidityInRange` itself at `0n`** with a new `[NEG_LIQ_IN_RANGE_GUARD]` (mirrors `[NEG_TLU_GUARD]` #856). This was the only in-range/reserve field in `updatePool` lacking a negative floor — reserves (#702), staked reserves (#771/#854), `totalLiquidityUSD` (#856) and `stakedLiquidityInRange` (#719) all clamp here. A Burn underflow (or tick-crossing drift) could otherwise drive it negative, corrupting snapshots / gauge-share **and** — as the staked clamp's ceiling — re-introducing a negative staked counter. Flooring the total first makes the staked clamp a faithful mirror of the reserve clamp (whose ceiling `clampedReserve0` is already `>= 0n`), so `min()` can never persist a negative staked value.
- **Log-level split** on the staked guard (mirrors #802): a zero total (pool pre-Swap/pre-init, or a Burn underflow floored to `0n`) is the benign #719 self-healing transient and logs at `info`; a genuine overshoot against a **populated** (`> 0n`) total — the #891 Superseed case — logs at `warn`.

The lower-bound `>= 0n` clamp (#719) is unchanged and still feeds the upper clamp.

## Test plan

Maps to the issue's acceptance criteria (`test/Aggregators/Pool.test.ts`):

- [x] After any update, `stakedLiquidityInRange <= liquidityInRange` holds — tracer test reproduces the Superseed `0xDaC2…` 2.7× overshoot (`1.96e15 → 7.28e14`); carry-forward + no-clamp cases covered.
- [x] Lower-bound `>= 0n` clamp (`[NEG_STAKED_LIQ_GUARD]`) still applies — three #719 tests stay green.
- [x] Upper-clamp guard log records pool, chain, pre-clamp value, `liquidityInRange`, clamped result — asserted field-by-field; `warn`/`info` level split asserted.
- [x] A test sets a staked replacement above `liquidityInRange` and asserts it clamps down — tracer test.
- [x] `liquidityInRange` floored to `0n` on Burn underflow + `[NEG_LIQ_IN_RANGE_GUARD]` logged — dedicated describe block; staked stays `>= 0n` as a consequence.
- [x] Full suite green (`vitest run`: 112 files, **1570 tests**), `tsc --noEmit` clean, `biome check` clean. CI `build-and-test` (install→codegen→build→qa→test) green.
- [ ] Flagged Superseed (5330) pools satisfy the invariant after a live reindex *(needs human — runtime reindex against chain 5330; guaranteed by construction + unit-reproduced here)*

## Code review notes

A multi-agent code review (43 agents, xhigh) surfaced three findings; all addressed:

- **Finding A (most severe) — negative staked.** The clamp's ceiling was un-floored, so a negative total persisted a negative `stakedLiquidityInRange`, defeating #719. **Fixed** by flooring `liquidityInRange` at the source (`[NEG_LIQ_IN_RANGE_GUARD]`); the staked ceiling is now always `>= 0n`, so `min()` of two non-negatives can't go negative. Re-verified SOUND by an independent adversarial pass. *(The floor on `liquidityInRange` was added in review follow-up — it also closes the latent negative-total data-quality gap, so AC #1 now holds unconditionally rather than only when the total is sane.)*
- **Finding C — log noise.** The guard logged unconditionally at `warn`. **Fixed** via the `warn`/`info` split above (mirrors #802).
- **Finding B — transient under-count (accepted tradeoff).** On gauge/NFPM/burn-only diffs that supply `stakedLiquidityInRange` but never refresh `liquidityInRange`, the clamp caps a correctly edge-derived staked value against the carried (possibly stale/zero) total, briefly understating staked liquidity. This is inherent to the AC-mandated approach: AC #1 requires `staked <= total` after **every** update, and the issue marks root-causing the staked-vs-total divergence as out of scope. It **self-heals on the next Swap** (when `liquidityInRange` is re-anchored absolutely from `slot0` and staked is re-derived) — the same property as the sibling #854/#702/#856 clamps. Caveat: an immutable `PoolSnapshot` taken inside the transient window records the understated value. A precise fix (re-deriving the total on non-Swap paths) is a possible follow-up beyond this issue's scope.

The two `liquidityInRange: 500n` seeds added to `CLStakedLiquidityEdgeSanity.test.ts` keep the #891 clamp from masking the orthogonal #719 derivation those tests verify — a staked position *is* part of the pool's total in-range liquidity, so seeding the total to cover it represents a physically valid state.

Closes #891

### DIFF
--- a/src/Aggregators/Pool.ts
+++ b/src/Aggregators/Pool.ts
@@ -454,6 +454,64 @@ export async function updatePool(
     );
   }
 
+  // Resolve the FINAL liquidityInRange persisted this update up-front so the
+  // upper-bound clamp below — and the object literal further down — compare
+  // against the exact value being written, not a stale prior. #703 semantics:
+  // an absolute diff.liquidityInRange (Swap-authoritative) wins; otherwise the
+  // incremental Mint/Burn delta lands on top of the carried value. (current
+  // and the increment are nullable, hence the `?? 0n` coalesces.)
+  const liquidityInRangeReplacement =
+    diff.liquidityInRange ??
+    (diff.incrementalLiquidityInRange ?? 0n) + (current.liquidityInRange ?? 0n);
+  // Clamp liquidityInRange to >= 0n (issue #891 follow-up). In-range liquidity
+  // is physically non-negative, but a Burn underflow (or accumulated
+  // tick-crossing drift) can drive the running value below zero. This was the
+  // only in-range/reserve field in updatePool lacking a floor — its siblings
+  // reserve0/1 (#702), stakedReserve0/1 (#771/#854), totalLiquidityUSD (#856)
+  // and stakedLiquidityInRange (#719) all clamp here. A negative total would
+  // also corrupt snapshots / gauge-share AND become a negative CEILING for the
+  // #891 staked clamp below (re-introducing a negative staked counter), so we
+  // floor it first — which keeps that clamp a faithful mirror of the reserve
+  // clamp, whose ceiling clampedReserve0 is already >= 0n. Clamp-and-log,
+  // mirroring [NEG_TLU_GUARD].
+  const finalLiquidityInRange =
+    liquidityInRangeReplacement < 0n ? 0n : liquidityInRangeReplacement;
+  if (liquidityInRangeReplacement < 0n) {
+    context.log.warn(
+      `[NEG_LIQ_IN_RANGE_GUARD][updatePool] field=liquidityInRange poolAddress=${current.poolAddress} chainId=${current.chainId} priorLiquidityInRange=${current.liquidityInRange} replacement=${liquidityInRangeReplacement} clampedTo=${finalLiquidityInRange}`,
+    );
+  }
+
+  // Upper-bound clamp: staked in-range liquidity is a SUBSET of the pool's
+  // total in-range liquidity and must never exceed it (issue #891). The staked
+  // value is derived from the staked-only tick-edge map (#719) while
+  // liquidityInRange is the Swap-authoritative total (#703); the two maps drift
+  // independently, so on tick-crossings the staked value can be left above the
+  // total (3 Superseed CL pools were observed in this state). Mirrors the
+  // [OVER_STAKED_RESERVE_GUARD] reserve clamp (#854): clamp
+  // stakedLiquidityInRange = min(stakedLiquidityInRange, liquidityInRange) and
+  // log the overshoot. `finalLiquidityInRange` is already floored >= 0n above
+  // (like the reserve clamp's clampedReserve0 ceiling), so the min() can never
+  // persist a negative staked value; the >= 0n lower clamp (#719) still applies.
+  const finalStakedLiquidityInRange =
+    clampedStakedLiquidityInRange > finalLiquidityInRange
+      ? finalLiquidityInRange
+      : clampedStakedLiquidityInRange;
+  if (clampedStakedLiquidityInRange > finalLiquidityInRange) {
+    // Log LEVEL split (mirrors #802's reserve-guard split): a zero total (pool
+    // pre-Swap/pre-init, or a Burn underflow floored to 0n above) is the benign
+    // #719 self-healing transient where a freshly-derived staked counter is
+    // briefly capped to 0n; log it at `info` so it doesn't flood the warn
+    // channel. A genuine overshoot against a POPULATED (> 0n) total — the #891
+    // Superseed case — stays at `warn` as a real-divergence tripwire.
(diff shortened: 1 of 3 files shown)


FonsSicca:
@coderabbitai review

